PR #63, opened 2022-07-13
Doc & deps refresh

chevdor:
It will be more convenient during dev to let cargo update the patch versions for external deps as needed and rely on semver to ensure nothing breaks but let me know @niklasad1 if there was another reason for fully specifying all versions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,14 +8,14 @@ edition = "2021"
 codec = { package = "parity-scale-codec", version = "3.0.0" }
 scale-info = { package = "scale-info", version = "2.1.1" }
 clap = { version = "3.2", features = ["derive", "env"] }
-tracing-subscriber = { version = "0.3.14", features = ["env-filter"]   }
+tracing-subscriber = { version = "0.3", features = ["env-filter"]   }
 jsonrpsee = { version = "0.14.0", features = ["ws-client", "macros"] }
-log = "0.4.17"
-paste = "1.0.7"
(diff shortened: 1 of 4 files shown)


niklasad1 (approved):
No particular reason, the repo was extracted from the staking miner in the polkadot repo.
So I haven't changed anything in the dependencies than to remove the unused ones.

I like this, thanks

However, I have changed so this repo can be used as a library as well so we can't rely entirely on `Cargo.lock`.

Cargo.toml
```diff
@@ -8,14 +8,14 @@ edition = "2021"
 codec = { package = "parity-scale-codec", version = "3.0.0" }
 scale-info = { package = "scale-info", version = "2.1.1" }
 clap = { version = "3.2", features = ["derive", "env"] }
-tracing-subscriber = { version = "0.3.14", features = ["env-filter"]   }
+tracing-subscriber = { version = "0.3", features = ["env-filter"]   }
```
niklasad1, line 11: we should keep this because https://github.com/tokio-rs/tracing/releases/tag/tracing-subscriber-0.3.13
fixed a compilation error in `version < 0.3.13`

```suggestion
tracing-subscriber = { version = "0.3.14", features = ["env-filter"]   }
```

chevdor:
> this repo can be used as a library as well so we can't rely entirely on Cargo.lock.

Good point. If other deps show as being problematic, we can tighen the versions again.

Issue #12, opened 2022-06-01
fetch weights from remote node


niklasad1:
Currently, the weights are hardcoded for each chain because the `MinerConfig trait` from substrate is required to be implemented.

Instead ideally it would be useful if we could just get the weights via runtime RPC API.
Perhaps something like:

```rust
   fn weight_from_extrinsic(uxt: UnsignedExtrinsic) -> Weight
```

Similar to https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/weights/pallet_election_provider_multi_phase.rs#L142-#L153

Then  just plug add the `Weight` as a parameter to the call to [mine_solution_with_snapshot](https://github.com/paritytech/substrate/blob/master/frame/election-provider-multi-phase/src/unsigned.rs#L400-#L404).

Which in turn requires changes in EPM pallet to remove `solution_weight` from the Miner trait.

That pretty much would make it work.

https://github.com/paritytech/substrate/issues/11665 might help with this.

/cc @kianenigma

kianenigma:
Optionally, `fn solution_weight` should stay, instead in the `impl MinerConfig for _` in the staking miner, instead we forward the call to the RPC. Might clash a bit with mixing `async` functions with a normal function, wdyt?

niklasad1:
it's a bit difficult to make `fn solution_weight` a future or something like that because the EPM pallet uses  that API and it needs to be implemented for the runtime IIRC. 

It's probably easier to remove some of the things from `MinerConfig` and try what I wrote above, it's a bit more work than I expected :P

niklasad1:
I read the code a bit closer now and I don't think it will work to use these introduced runtime RPC APIs will work for our use-case because it's used for an extrinsic and in our case we need to have the `RawSolution` to get the weight from the extrinsic "submit_unsigned". The `WeightInfo` is used to create the `RawSolution` so it's a chicken and egg problem, thus we would need some RPC interfaces to get the weight info based on `voters, targets, active voters, desired winner` from the snapshot.

Otherwise, we would have to kill the  `MinerConfig` because of https://github.com/paritytech/substrate/blob/master/frame/election-provider-multi-phase/src/unsigned.rs#L636-#L638  and that kills the entire purpose. 

As I see it:

1. make another hack with four global variables (voters, targets, active voters, desired winner) which are set every time a solution is mined and then `fn solution_weight` would read that instead of the hardcoded values.
[...]

PR #237, opened 2022-08-23
fetch weight from remote node

niklasad1:
Closing #12

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -138,17 +138,16 @@ pub mod westend {
 		type Solution = NposSolution16;
 
 		// SYNC
-		fn solution_weight(voters: u32, targets: u32, active_voters: u32, degree: u32) -> Weight {
+		fn solution_weight(voters: u32, targets: u32, active_voters: u32, _degree: u32) -> Weight {
 			use pallet_election_provider_multi_phase::{RawSolution, SolutionOrSnapshotSize};
 
-			let raw = {
-				let solution = NposSolution16 {
+			// Mock a RawSolution to get the correct weight wihout having to do the heavy work.
+			let raw = RawSolution {
+				solution: NposSolution16 {
 					votes1: mock_votes(active_voters),
 					..Default::default()
-				};
-				let mut raw = RawSolution::default();
-				raw.solution = solution;
-				raw
+				},
+				..Default::default()
 			};
 
 			let tx = runtime::tx()
@@ -227,17 +226,16 @@ pub mod polkadot {
 		type Solution = NposSolution16;
 
 		// SYNC
-		fn solution_weight(voters: u32, targets: u32, active_voters: u32, degree: u32) -> Weight {
+		fn solution_weight(voters: u32, targets: u32, active_voters: u32, _degree: u32) -> Weight {
 			use pallet_election_provider_multi_phase::{RawSolution, SolutionOrSnapshotSize};
 
-			let raw = {
-				let solution = NposSolution16 {
+			// Mock a RawSolution to get the correct weight wihout having to do the heavy work.
+			let raw = RawSolution {
+				solution: NposSolution16 {
 					votes1: mock_votes(active_voters),
 					..Default::default()
-				};
-				let mut raw = RawSolution::default();
-				raw.solution = solution;
-				raw
+				},
+				..Default::default()
 			};
 
 			let tx = runtime::tx()
@@ -316,17 +314,16 @@ pub mod kusama {
 		type Solution = NposSolution24;
 
 		// SYNC
-		fn solution_weight(voters: u32, targets: u32, active_voters: u32, degree: u32) -> Weight {
+		fn solution_weight(voters: u32, targets: u32, active_voters: u32, _degree: u32) -> Weight {
 			use pallet_election_provider_multi_phase::{RawSolution, SolutionOrSnapshotSize};
 
-			let raw = {
-				let solution = NposSolution24 {
+			// Mock a RawSolution to get the correct weight wihout having to do the heavy work.
+			let raw = RawSolution {
+				solution: NposSolution24 {
 					votes1: mock_votes(active_voters),
 					..Default::default()
-				};
-				let mut raw = RawSolution::default();
-				raw.solution = solution;
-				raw
+				},
+				..Default::default()
 			};
 
 			let tx = runtime::tx()
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -12,7 +12,7 @@ use once_cell::sync::OnceCell;
 use pallet_transaction_payment::RuntimeDispatchInfo;
 use sp_core::Bytes;
 
-pub static SHARED_URI: OnceCell<String> = OnceCell::new();
+pub static SHARED_CLIENT: OnceCell<SubxtClient> = OnceCell::new();
 
 macro_rules! impl_atomic_u32_parameter_types {
 	($mod:ident, $name:ident) => {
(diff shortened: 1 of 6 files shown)


src/chain.rs
```diff
@@ -333,3 +377,41 @@ pub mod kusama {
...
+		let info: RuntimeDispatchInfo<Balance> = Decode::decode(&mut bytes.0.as_ref()).unwrap();
+
+		log::debug!(target: LOG_TARGET, "Received weight for remote note: {:?}", info.weight);
+
+		info.weight
+	})
+}
+
+fn mock_votes(voters: u32) -> Vec<(u32, u16)> {
+	(0..voters).map(|v| (v, 0)).collect()
```
niklasad1, line 416: /cc @kianenigma is this ok?

I use the target index as `0` for every vote.

src/chain.rs
```diff
@@ -6,6 +6,14 @@
 // polkadot, that only has `const` and `type`s that are used in the runtime, and we can import
 // that.
 
+use codec::{Decode, Encode};
+use jsonrpsee::{core::client::ClientT, rpc_params};
+use once_cell::sync::OnceCell;
+use pallet_transaction_payment::RuntimeDispatchInfo;
+use sp_core::Bytes;
+
+pub static SHARED_CLIENT: OnceCell<SubxtClient> = OnceCell::new();
```
niklasad1, line 15: this is needed because the `MinerConfig trait` doesn't take &self, so I had to use a global variable fine for now.

src/monitor.rs
```diff
@@ -115,10 +115,12 @@ macro_rules! monitor_cmd_for {
 						return;
 					}
 
-					let addr = runtime::storage().election_provider_multi_phase().round();
-					let round = match api.storage().fetch(&addr, Some(hash)).await {
+					let round = match api.storage().fetch(&runtime::storage().election_provider_multi_phase().round(), Some(hash)).await {
 						Ok(Some(round)) => round,
-						Ok(None) => unreachable!("Round must always exist"),
+						Ok(None) => {
+							log::warn!(target: LOG_TARGET, "EPM::Round was not found at: {:?}", at.hash());
```
niklasad1, line 121: I have seen this on my dev node when no signed phase occured yet.

src/monitor.rs
```diff
@@ -219,7 +221,10 @@ macro_rules! monitor_cmd_for {
 					match res {
 						Ok(Some(Phase::Signed)) => Ok(()),
 						Ok(Some(_)) => Err(Error::IncorrectPhase),
-						Ok(None) => unreachable!("Phase should always exist"),
+						Ok(None) => {
+							log::warn!(target: LOG_TARGET, "EPM::Phase was not found at: {:?}", hash);
```
niklasad1, line 225: I have seen this on my dev node when no signed phase occured yet.

src/main.rs
```diff
@@ -62,6 +62,8 @@ async fn main() -> Result<(), Error> {
 		.map_err(|e| log::warn!("Failed to start prometheus endpoint: {}", e));
 	log::info!(target: LOG_TARGET, "Connected to chain: {}", chain);
 
+	chain::SHARED_CLIENT.set(api.clone()).expect("shared URI only set once; qed");
```
niklasad1, line 65:
```suggestion
	chain::SHARED_CLIENT.set(api.clone()).expect("shared client only set once; qed");
```

src/chain.rs
```diff
@@ -282,16 +314,23 @@ pub mod kusama {
...
-			// feasibility weight.
-			(31_722_000 as Weight)
-				// Standard Error: 8_000
-				.saturating_add((1_255_000 as Weight).saturating_mul(v as Weight))
-				// Standard Error: 28_000
-				.saturating_add((8_972_000 as Weight).saturating_mul(a as Weight))
-				// Standard Error: 42_000
-				.saturating_add((966_000 as Weight).saturating_mul(d as Weight))
-				.saturating_add(static_types::DbWeight::get().reads(4 as Weight))
+		fn solution_weight(voters: u32, targets: u32, active_voters: u32, _degree: u32) -> Weight {
```
kianenigma, line 317: d is not degree -- it is `desired_targets`, i.e. the number of validators.

src/chain.rs
```diff
@@ -333,3 +377,44 @@ pub mod kusama {
...
+
+		let info: RuntimeDispatchInfo<Balance> = Decode::decode(&mut bytes.0.as_ref()).unwrap();
+
+		log::debug!(target: LOG_TARGET, "Received weight for remote note: {:?}", info.weight);
+
+		info.weight
+	})
+}
+
+fn mock_votes(voters: u32) -> Vec<(u32, u16)> {
```
kianenigma, line 418: No -- this function should also take `desired_targets` aka `d` as input and make sure that there are `d` unqiue unique targets, not just `0`.

src/chain.rs
```diff
@@ -282,16 +314,23 @@ pub mod kusama {
...
-				.saturating_add((8_972_000 as Weight).saturating_mul(a as Weight))
-				// Standard Error: 42_000
-				.saturating_add((966_000 as Weight).saturating_mul(d as Weight))
-				.saturating_add(static_types::DbWeight::get().reads(4 as Weight))
+		fn solution_weight(voters: u32, targets: u32, active_voters: u32, _degree: u32) -> Weight {
+			use pallet_election_provider_multi_phase::{RawSolution, SolutionOrSnapshotSize};
+
+			// Mock a RawSolution to get the correct weight wihout having to do the heavy work.
+			let raw = RawSolution {
+				solution: NposSolution24 {
```
kianenigma, line 322: here, once you build this fake solution, you should debug assert that: 

```
assert_eq!(solution.voter_count(), a);
assert_eq!(solution.unique_targets(), desired_targets);
```

and then you can be sure that you will get the correct weight.

src/monitor.rs
kianenigma: Yeah once the first round has not even started, this will be zero.

src/monitor.rs
kianenigma: Seems like subxt is not capable of parsing the defaults that we define for storage types. 

https://github.com/paritytech/substrate/blob/49b06901eb65f2c61ff0934d66987fd955d5b8f5/frame/election-provider-multi-phase/src/lib.rs#L1193

The `default` for `Phase` is `None`. You can just hardcode this for now.

kianenigma: Similarly, the default for `Round` is 1:

https://github.com/paritytech/substrate/blob/49b06901eb65f2c61ff0934d66987fd955d5b8f5/frame/election-provider-multi-phase/src/lib.rs#L1188

kianenigma (changes requested):
One mistake with `desired_targets` otherwise looks good.

src/chain.rs
niklasad1: yes sir, good idea

I think we can have `assert_eq` there and not debug assert to catch those in the "burn-in"

src/chain.rs
```diff
@@ -333,3 +413,55 @@ pub mod kusama {
...
+			"Received weight of `Solution Extrinsic` from remote node: {:?}",
+			info.weight
+		);
+
+		info.weight
+	})
+}
+
+fn mock_votes(voters: u32, desired_targets: u16) -> Vec<(u32, u16)> {
+	assert!(voters >= desired_targets as u32);
```
niklasad1, line 459: @kianenigma 

I guess this must be true otherwise won't "all desired targets" be used

src/chain.rs
```diff
@@ -130,16 +138,35 @@ pub mod westend {
...
+					votes1: mock_votes(
+						active_voters,
+						desired_targets.try_into().expect("Desired targets < u16::MAX"),
+					),
+					..Default::default()
+				},
+				..Default::default()
+			};
+
+			assert_eq!(raw.solution.voter_count(), active_voters as usize);
```
kianenigma, line 162: As long as these pass, we're good.

kianenigma (approved):
Bueno

PR #283, opened 2022-09-06
Fix `--file` reference for `buildah` in CI

rcny:


sergejparity (approved):

radupopa2010 (commented):
Are you sure this will work?

line 163:
```
    - mv Dockerfile artifacts/
```

rcny:
Why not? See [this](https://github.com/paritytech/staking-miner-v2/blob/main/.gitlab-ci.yml#L187) line.

PR #322, opened 2022-09-14
fix: update metadata on upgrades

niklasad1:
Currently when the miner starts up it fetches the constants that are needed and subxt compares these "to the metadata on the connected node" but on runtime upgrades those are never updated these which this PR fixes.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -116,19 +116,23 @@ macro_rules! helpers_for_runtime {
 				Ok((voters, targets, desired_targets))
 			}
 
-			pub fn [<update_runtime_constants_$runtime>](api: &SubxtClient) {
+			pub fn [<update_runtime_constants_$runtime>](api: &SubxtClient) -> Result<(), Error> {
 				use chain::static_types;
 				use sp_runtime::Perbill;
 				use crate::chain::[<$runtime>]::runtime;
 
 				// maximum weight of the signed submission is exposed from metadata and MUST be this.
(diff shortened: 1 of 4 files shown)


kianenigma (approved):

src/helpers.rs
```diff
@@ -116,19 +116,23 @@ macro_rules! helpers_for_runtime {
...
 
-			pub fn [<update_runtime_constants_$runtime>](api: &SubxtClient) {
+			pub fn [<update_runtime_constants_$runtime>](api: &SubxtClient) -> Result<(), Error> {
 				use chain::static_types;
 				use sp_runtime::Perbill;
 				use crate::chain::[<$runtime>]::runtime;
 
 				// maximum weight of the signed submission is exposed from metadata and MUST be this.
-				let max_weight = api.constants().at(&runtime::constants().election_provider_multi_phase().signed_max_weight()).expect("constant `max weight` must exist").ref_time;
+				let max_weight = api.constants().at(&runtime::constants().election_provider_multi_phase().signed_max_weight())
```
kianenigma, line 125: When this is called after a runtime upgrade again, will it actually load new constant? won't it need a new `.scale` file, and therefore just load the same thing?

kianenigma: this would also make a good integration test: upgrade the runtime to have less weight or something like that, and then check it is actually reflected here.

niklasad1: I think need to load the same thing otherwise it will fail (need a new scale file) but I need to test it with a runtime upgrade to make sure.

niklasad1:
This doesn't work properly we need https://github.com/paritytech/subxt/issues/652 or provide the runtime subscription ourselves.

PR #324, opened 2022-09-14
metrics for weight and length

kianenigma:
- [x] weight and length metrics


src/monitor.rs
```diff
@@ -141,12 +144,39 @@ macro_rules! monitor_cmd_for {
...
 						(Ok((solution, score, size)), elapsed) => {
 							let elapsed_ms = elapsed.as_millis();
-							crate::prometheus::observe_mined_solution_duration(elapsed_ms as f64);
-							crate::prometheus::set_score(score);
-							log::trace!(target: LOG_TARGET, "Mined solution with {:?} size: {:?} round: {:?} at: {}, took: {} ms", score, size, round, at.number(), elapsed_ms);
+							let encoded_len = solution.encoded_size();
+							let active_voters = solution.voter_count() as u32;
+							let desired_targets = solution.unique_targets().len() as u32;
+
+							let final_weight = tokio::task::spawn_blocking(move || {
```
niklasad1, line 154: @kianenigma this will spawn the on task on the blocking thread pool in the tokio

if we call it on the `async scheduler` it will fuck up the scheduler i.e, in `tokio::task` so this should work now.

niklasad1: it basically hack around that we are forced to block the current thread because of `MinerConfig::solution_weight`

niklasad1 (approved):

src/prometheus.rs
```diff
@@ -192,6 +193,22 @@ mod hidden {
 		))
 		.unwrap()
 	});
+	static SUBMISSION_LENGTH: Lazy<Gauge> = Lazy::new(|| {
+		register_gauge!(opts!(
+			"staking_miner_solution_length_byes",
```
niklasad1, line 198:
```suggestion
			"staking_miner_solution_length_bytes",
```

src/chain.rs
```diff
@@ -464,6 +463,7 @@ fn mock_votes(voters: u32, desired_targets: u16) -> Vec<(u32, u16)> {
 }
 
 #[cfg(test)]
+#[test]
```
niklasad1, line 466: :facepalm:

niklasad1: nice catch

niklasad1:
@kianenigma merge this? looks great

PR #387, opened 2022-09-29
WIP: kill macros + one metadata file

niklasad1:
Builds on top of #357


src/epm_dynamic.rs
```diff
@@ -0,0 +1,213 @@
...
+	decode_as_type(&mut bytes.as_ref(), ty_id, &types)
+		.map(|v| v.remove_context())
+		.map_err(|e| {
+			Error::DynamicTransaction(format!(
+				"Failed to decode {}: {:?}",
+				std::any::type_name::<T>(),
+				e
+			))
+		})
+}
```
jsdw, line 152: Ooh, I see, this is a clever way to take some type implementing `TypeInfo + Encode` into a `scale_value::Value`! It's a shame to have to encode it to bytes and then decode from the bytes, but offhand I can't think of a better/more reliable way.

You _can_ go from a `T: Serialize` straight into a `scale_value::Value`, which would be more efficient (no intermediate step), but it also goes through the serde data model. Might be worth trying if you wanted faster and fewer lines of code, but the downside is that the serde and scale impls don't necessarily need to agree with eachother (eg H256 serializes to a string, which would turn into a scale_value primitive String rather than a sequence of bytes, and the string then wouldn't encode to bytes as would be needed to encode it properly).

src/epm_dynamic.rs
```diff
@@ -0,0 +1,213 @@
...
+
+	encode_as_type(val, ty_id, &types, &mut bytes).map_err(|e| {
+		Error::DynamicTransaction(format!(
+			"Failed to encode {}: {:?}",
+			std::any::type_name::<T>(),
+			e
+		))
+	})?;
+	Ok(bytes)
+}
```
jsdw, line 167: If you wanted to pass this to some call (`api.tx().submit..`) as a `Value`, this would use the runtime metadata to encode properly I'd hope. But I'm guessing you ran into a problem or couldn't do that so I'm intrigued :)

niklasad1: I had to use this because I wanted to read a storage value and then `codec::Decode::decode -> SignedSubmission` see `fn signed_submission_at` above.

The reason is that the `epm pallet` types doesn't implement serde, not sure if there is a more clever way to do it :)

src/epm_dynamic.rs
niklasad1: yeah but these types doesn't implement `serde` maybe I can convince @kianenigma to add  `deserialize/serialize` on these types :)

jsdw: Makes sense! Aside from using Serde, or some other similar machinery, I think what you've done is the best possible approach anyways :)

src/helpers.rs
```diff
@@ -50,114 +51,86 @@ pub trait TimedFuture: Sized + Future {
...
+		Ok(Err(err)) => Err(Error::Other(format!("{:?}", err))),
+		Err(err) => Err(Error::Other(format!("{:?}", err))),
+	}
+}
 
-				static_types::DbWeight::set(system_db_weight);
-				static_types::MaxWeight::set(max_weight);
-				static_types::MaxLength::set(max_length);
-				static_types::MaxVotesPerVoter::set(max_votes_per_voter);
+pub async fn snapshot<T: MinerConfig + Send + Sync + 'static>(
```
kianenigma, line 93: I am generally inclined to ask you to document all public functions in here. Some are clear, some are not. And you and I both know this codebase fairly well, which means we're biased.

src/helpers.rs
```diff
@@ -50,114 +51,86 @@ pub trait TimedFuture: Sized + Future {
...
+		.await?
+		.unwrap_or_default();
+
+	let desired_targets = api
+		.storage()
+		.fetch(&runtime::storage().election_provider_multi_phase().desired_targets(), hash)
+		.await?
+		.unwrap_or_default();
+
+	let voters: Vec<_> = voters
```
kianenigma, line 113: I recall I once looked at why we need this hacky code and it seemed neccessary, I want to revisit it now again.

src/prelude.rs
```diff
@@ -59,4 +62,43 @@ pub type Accuracy = sp_runtime::Perbill;
...
+	AccountId,
+	VoteWeight,
+	frame_support::BoundedVec<AccountId, static_types::MaxVotesPerVoter>,
+)>;
+pub type Snapshot = (BoundedVoters, Vec<AccountId>, u32);
+pub type SignedSubmission<S> =
+	pallet_election_provider_multi_phase::SignedSubmission<AccountId, Balance, S>;
+
+#[subxt::subxt(
+	runtime_metadata_path = "artifacts/metadata.scale",
```
kianenigma, line 80: would be good to have a note in the README about how to acquire this.

kianenigma (approved):
Would like to clone to code and peek around myself as well, but certainly LGTM.

src/helpers.rs
niklasad1: I tried to write some docs for these and move the epm related stuff to the epm module instead.

src/helpers.rs
niklasad1: it's to fetch to the snapshot data as we use `Miner::mine_solution_with_snapshot` added some docs.

PR #402, opened 2022-10-28
companion for the recent epm changes

niklasad1:
This PR fixes such that `staking-miner-v2` works against polkadot master again.


src/monitor.rs
```diff
@@ -334,31 +334,31 @@ where
...
 	let addr = runtime::storage().election_provider_multi_phase().signed_submission_indices();
 	let indices = api.storage().fetch_or_default(&addr, Some(at)).await?;
 
 	log::debug!(target: LOG_TARGET, "submitted solutions: {:?}", indices.0);
 
-	for (other_score, _) in indices.0 {
-		if !score.strict_threshold_better(other_score, epsilon) {
-			return Err(Error::BetterScoreExist)
-		}
+	if indices
```
niklasad1, line 353: //cc @kianenigma you added [this check](https://github.com/paritytech/polkadot/blob/6c3d5add1c53f3daff6800420c7e65a18a0f0ed1/utils/staking-miner/src/monitor.rs#L123-#L126)

I missed it in the review but why is that needed? AFAIU, the solution is better than solutions is in queue according the "strategy" and if that's ttrue then the solution with worse will be kicked out?

kianenigma: yeah I checked the original PR as well and it doesn't make sense, the whole `if` has to go away. Can you make sure it is fixed in Polkadot as well?

kianenigma (approved):

src/monitor.rs
niklasad1: sure

PR #417, opened 2022-11-16
chore: update deps

niklasad1:


tests/monitor.rs
```diff
@@ -6,7 +6,6 @@ pub mod common;
 use assert_cmd::cargo::cargo_bin;
 use codec::Decode;
 use common::{init_logger, run_polkadot_node, KillChildOnDrop};
-use pallet_election_provider_multi_phase::{ElectionCompute, ReadySolution};
```
niklasad1, line 9: some new trait bounds in substrate broke this, let's use subxt types for now

PR #425, opened 2022-12-01
fix: `--listen finalized` wait for finalized xt

niklasad1:
When one runs `--listen finalized` it's possible for the block not to be finalized when it's included in block.

Thus, when that task finishes it's possible to that the `next task` will not see that submission as it is not finalized yet and submit more than one submission in a round. 

Should fix the duplicate solutions that I spotted on westend.

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -316,9 +316,16 @@ async fn mine_and_submit_solution<T>(
 	};
 
 	prometheus::on_submission_attempt();
-	match submit_and_watch_solution::<T>(&api, signer, (solution, score, round), hash, nonce)
-		.timed()
-		.await
+	match submit_and_watch_solution::<T>(
+		&api,
+		signer,
+		(solution, score, round),
+		hash,
+		nonce,
+		config.listen,
+	)
+	.timed()
+	.await
 	{
 		(Ok(_), now) => {
 			prometheus::on_submission_success();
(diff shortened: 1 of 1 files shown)


src/monitor.rs
```diff
@@ -448,11 +456,13 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
 						.find_first::<runtime::election_provider_multi_phase::events::SolutionStored>(
 						);
 
-				return if let Ok(Some(_)) = solution_stored {
+				if let Ok(Some(_)) = solution_stored {
 					log::info!(target: LOG_TARGET, "Included at {:?}", details.block_hash());
-					Ok(())
+					if let Listen::Head = listen {
```
niklasad1, line 461: if one "--listen head` it's fine to regard it as completed here but not in when `--listen finalized`

this is done to avoid submitting duplicate solutions by mistake

src/monitor.rs
```diff
@@ -448,11 +456,13 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
 						.find_first::<runtime::election_provider_multi_phase::events::SolutionStored>(
 						);
```
kianenigma, line 457: formatting seems broken here?

niklasad1: rustfmt wants it like that :sob:

PR #443, opened 2022-12-19
fix: `dryRun check` + `fix multiple solutions bug`

niklasad1:


src/monitor.rs
```diff
@@ -292,6 +292,20 @@ async fn mine_and_submit_solution<T>(
 		return
 	}
 
+	if let Err(e) =
+		ensure_no_previous_solution::<T::Solution>(&api, best_head, signer.account_id()).await
```
niklasad1, line 296: //cc @kianenigma we only checked this when the block was received (i.e, heads subscriptions) and after solution was mined we need to ensure that the account hasn't submitted at the latest state.

src/monitor.rs
```diff
@@ -426,6 +440,14 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
 		.tx()
 		.create_signed_with_nonce(&tx, &*signer, nonce, ExtrinsicParams::default())?;
 
+	let outcome = api.rpc().dry_run(xt.encoded(), None).await?;
```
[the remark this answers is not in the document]
kianenigma: Should be optional, via a flag, in case we are running against a node that does not support unsafe RPCs. 

Perhaps the default can be with dry-run, and `--no-dry-run` flag added.

Or, if you prefer to keep it flexible, we always do-run iff it is available

Lastly, needs to inform devops that our nodes should expose unsafe RPCs :)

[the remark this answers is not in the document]
niklasad1: let's add a flag then forgot that that RPC is unsafe

src/monitor.rs
```diff
@@ -130,11 +130,15 @@ fn kill_main_task_if_critical_err(tx: &tokio::sync::mpsc::UnboundedSender<Error>
...
 							const VERIFICATION_ERROR: i32 = 1002;
+							use jsonrpsee::types::error::ErrorCode;
 
 							// Check if the transaction gets fatal errors from the `author` RPC.
 							// It's possible to get other errors such as outdated nonce and similar
 							// but then it should be possible to try again in the next block or round.
-							if e.code() == BAD_EXTRINSIC_FORMAT || e.code() == VERIFICATION_ERROR {
+							if e.code() == BAD_EXTRINSIC_FORMAT ||
+								e.code() == VERIFICATION_ERROR || e.code() ==
+								ErrorCode::MethodNotFound.code()
```
niklasad1, line 140: if we got back `method not found` is most likely because the RPC method was unsafe, then just terminate.

all solutions will be rejected anyway then

kianenigma:
Please add these project to the NPoS project as it helps me find them faster 👍

src/opt.rs
```diff
@@ -223,6 +223,13 @@ pub struct MonitorConfig {
 	/// configured, it might re-try and lose funds through transaction fees/deposits.
 	#[clap(long, short, env = "SEED")]
 	pub seed_or_path: String,
+
+	/// Verify the submission by `dry-run` the extrinsic to check the validity.
+	/// If it fails then the block is ignored.
```
kianenigma, line 228: What does this mean? What block is ignored? 

I guess what you are trying to say is: If this is set, against a node that does not support unsafe RPCs, then we don't submit anything? 

I think we should test this at startup, and immediately terminate. Every time we connect to a node, if we have this flag set, we should check it it supports the unsafe stuff or not.

kianenigma (changes requested):
See: https://github.com/paritytech/staking-miner-v2/pull/443/files#r1063275125

src/opt.rs
niklasad1: Nah, it means if the dry-run fails on a solution at block x then we simply emit a warning and drop that solution.
Then we try on the next block.

Fair enough, it could be good to have a proper error if the RPC call itself is denied because of "unsafe", currently we just terminate when that occurs with error from substrate.

kianenigma (approved):

PR #450, opened 2023-01-11
chore: release v0.1.2

niklasad1:


kianenigma (approved):
Seeing the nice changelog, I want to do a brain dump about deployment and community: 

Long term, We should have a much more extensive README with the end goal that anyone interested in running this tool would land in it, and learns about: 

1. how to run this.
2. the risks associated. 

I think the situation with `dry-run` was a good lesson that the staking miner should have a set of strict guarantees, all noted in the README, and it should do its best to adhere to them. IMO, the north star should be this: "Running this might not bring you benefit, but it will not lose money either".

PR #462, opened 2023-02-14
feat: add prometheus metric `submission_interval`

### LINKED ISSUE #459: Have network specific alerts for staking-miner-v2
We have alert if staking miner did not trigger any submissions:
```
alert:StakingMinerNoSubmissions 
expr:rate(staking_miner_submissions_started[5m]) == 0
for: 25h         <= session  time  + 1h 
```

Since session time is different for each network (e.g westend/kusama/polkadot) this alert will be correct for polkadot 24h but not for westend 6h 

Is it possible to expose session length as metrics?

niklasad1:
Close #459 

Each chain has different configurations how often an election occurs and this PR adds functionality to read the associated constants to calculate it.

Staking miner will update these on startup and on each runtime upgrade.

### DIFF
--- a/src/epm.rs
+++ b/src/epm.rs
@@ -55,19 +55,21 @@ pub(crate) async fn update_metadata_constants(api: &SubxtClient) -> Result<(), E
 	const SIGNED_MAX_WEIGHT: Constant = Constant::new(EPM_PALLET_NAME, "SignedMaxWeight");
 	const MAX_LENGTH: Constant = Constant::new(EPM_PALLET_NAME, "MinerMaxLength");
 	const MAX_VOTES_PER_VOTER: Constant = Constant::new(EPM_PALLET_NAME, "MinerMaxVotesPerVoter");
+	const EPOCH_DURATION: Constant = Constant::new("Babe", "EpochDuration");
+	const SESSIONS_PER_ERA: Constant = Constant::new("Staking", "SessionsPerEra");
 
 	// TODO: no constant is exposed directly for this, hardcoded for now.
 	const SLOT_DURATION_SECS: u32 = 6;
-	const EPOCH_DURATION: Constant = Constant::new("Babe", "EpochDuration");
-	const SESSIONS_PER_ERA: Constant = Constant::new("Staking", "SessionsPerEra");
 
 	let max_weight = read_constant::<Weight>(api, SIGNED_MAX_WEIGHT)?;
 	let max_length: u32 = read_constant(api, MAX_LENGTH)?;
 	let max_votes_per_voter: u32 = read_constant(api, MAX_VOTES_PER_VOTER)?;
-	let epoch: u32 = read_constant(api, EPOCH_DURATION)?;
-	let sessions: u32 = read_constant(api, SESSIONS_PER_ERA)?;
-	// era = epoch * sessions * slot
-	let era_hours: u32 = (epoch * sessions * SLOT_DURATION_SECS) / (60 * 60);
+	// session == epoch
+	let session: u32 = read_constant(api, EPOCH_DURATION)?;
+	let num_sessions_per_era: u32 = read_constant(api, SESSIONS_PER_ERA)?;
+
+	// era = session * seconds per slot * number sessions per era
+	let era_hours: u32 = (session * SLOT_DURATION_SECS * num_sessions_per_era) / (60 * 60);
 
 	log::trace!(
 		target: LOG_TARGET,
--- a/src/epm.rs
+++ b/src/epm.rs
@@ -16,7 +16,7 @@
 
 //! Wrappers or helpers for [`pallet_election_provider_multi_phase`].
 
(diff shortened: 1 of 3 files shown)


src/prometheus.rs
```diff
@@ -161,6 +161,13 @@ mod hidden {
 		register_gauge!(opts!("staking_miner_solution_weight", "Weight of the solution submitted"))
 			.unwrap()
 	});
+	static ERA: Lazy<Gauge> = Lazy::new(|| {
+		register_gauge!(opts!(
+			"staking_miner_era_hours",
```
niklasad1, line 166: @BulatSaif any preferred naming of this?

You wrote in the issue session length but I guess we want the actual "Era period/interval" here

src/epm.rs
```diff
@@ -31,34 +31,45 @@ use subxt::{dynamic::Value, rpc::rpc_params, tx::DynamicTxPayload};
...
 
 	let max_weight = read_constant::<Weight>(api, SIGNED_MAX_WEIGHT)?;
 	let max_length: u32 = read_constant(api, MAX_LENGTH)?;
 	let max_votes_per_voter: u32 = read_constant(api, MAX_VOTES_PER_VOTER)?;
+	// session == epoch
+	let session: u32 = read_constant(api, EPOCH_DURATION)?;
+	let num_sessions_per_era: u32 = read_constant(api, SESSIONS_PER_ERA)?;
+
+	// era = session * seconds per slot * number sessions per era
+	let era_hours: u32 = (session * SLOT_DURATION_SECS * num_sessions_per_era) / (60 * 60);
```
niklasad1, line 72: @kianenigma can you double check this?

Not really familar with this stuff :P

src/epm.rs
```diff
@@ -31,34 +31,46 @@ use subxt::{dynamic::Value, rpc::rpc_params, tx::DynamicTxPayload};
...
-	const MAX_LENGTH: EpmConstant = EpmConstant::new("MinerMaxLength");
-	const MAX_VOTES_PER_VOTER: EpmConstant = EpmConstant::new("MinerMaxVotesPerVoter");
+	const SIGNED_MAX_WEIGHT: Constant = Constant::new(EPM_PALLET_NAME, "SignedMaxWeight");
+	const MAX_LENGTH: Constant = Constant::new(EPM_PALLET_NAME, "MinerMaxLength");
+	const MAX_VOTES_PER_VOTER: Constant = Constant::new(EPM_PALLET_NAME, "MinerMaxVotesPerVoter");
+	const EPOCH_DURATION: Constant = Constant::new("Babe", "EpochDuration");
+	const SESSIONS_PER_ERA: Constant = Constant::new("Staking", "SessionsPerEra");
+
+	// TODO: no constant is exposed directly for this, hardcoded for now.
+	const SLOT_DURATION_SECS: u32 = 6;
```
niklasad1, line 62: //cc @kianenigma I only found [this](https://github.com/paritytech/polkadot/blob/master/runtime/westend/constants/src/lib.rs#L44) which is exposed indirectly via [MinimumPeriod](https://github.com/paritytech/polkadot/blob/master/runtime/westend/src/lib.rs#L288) which isn't what I want...

Thus, hardcoded it

src/epm.rs
```diff
@@ -31,34 +31,46 @@ use subxt::{dynamic::Value, rpc::rpc_params, tx::DynamicTxPayload};
...
 
 	let max_weight = read_constant::<Weight>(api, SIGNED_MAX_WEIGHT)?;
 	let max_length: u32 = read_constant(api, MAX_LENGTH)?;
 	let max_votes_per_voter: u32 = read_constant(api, MAX_VOTES_PER_VOTER)?;
+	// session == epoch
+	let session: u32 = read_constant(api, EPOCH_DURATION)?;
+	let num_sessions_per_era: u32 = read_constant(api, SESSIONS_PER_ERA)?;
+
+	// era = session * seconds per slot * number sessions per era
+	let submission_interval_hours: u32 =
```
kianenigma, line 72: This number is more or less the same all the time FYI, perhaps we can communicate it to the monitoring system in some other way other than pushing it to prometheus?

kianenigma (commented):
Code is fine, but I am not sure why this needs to be a "metric", as in "metric that can change over time". This value is _constant per chain_.

src/epm.rs
BulatSaif: Sure we can set this value directly in monitoring, I thought it will be good to have a universal alert for any chains (who knows how many new chains we will have). But if need to hardcode some value for monitoring in code it is better to do it on the monitoring side (it is more flexible, we don't need to wait for the release to change it)

niklasad1: It's not hardcoded (except one thing that I will fix) but in practice these values "constants" as Kian said.

However, as we read it from the metadata/storage on each startup/runtime upgrade it's probably better than to handle it in monitoring.

PR #538, opened 2023-04-26
add trimming tests

niklasad1:
Basically it works like this:

1. Compile the staking-miner-playground (--features test-trimming) that will ensure that both MinerMaxLength and MinerMaxWeight is configured to a smaller value and trimming always occurs.

For instance recall that Kian added some special code where the number of validators, nominators etc can be configured by env variables and we read the same env variables to ensure that MaxLength is less than the number of validators.

2. Run the staking miner and assert:
    a) Trimming weight and length from the CLI output isn't 0
    b) the solution is accepted by chain (hence feasibility check was successful)


staking-miner-playground/runtime/src/lib.rs
```diff
@@ -8,6 +8,17 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 pub mod pallet_config_block;
 
+#[macro_export]
+macro_rules! prod_or_test {
+	($prod:expr, $test:expr) => {
+		if cfg!(feature = "test-trimming") {
```
niklasad1, line 14: the staking miner playground must be compiled with `--features test-trimming`

staking-miner-playground/runtime/src/lib.rs
```diff
@@ -450,15 +461,21 @@ parameter_types! {
...
-		<Runtime as frame_system::Config>::BlockWeights::get()
-		.get(DispatchClass::Normal)
-		.max_total
-		.unwrap();
-	pub MinerMaxLength: u32 = Perbill::from_rational(8u32, 10) *
-		*(<<Runtime as frame_system::Config>::BlockLength as Get<limits::BlockLength>>::get())
-		.max
-		.get(DispatchClass::Normal);
+
+	// TODO(niklasad1): this is a hack to get the number of validators, candidates and nominators
```
niklasad1, line 465: this question is really how configure these such that it doesn't too much such that `Incorrect number winners` doesn't occur

//cc @gpestana @kianenigma

tests/monitor.rs
```diff
@@ -4,82 +4,87 @@
...
+	assert!(has_trimming_output(miner).await);
+}
+
+// Helper that parses the CLI output to find logging outputs based on the following:
+//
+// i) DEBUG runtime::election-provider: 🗳 from 934 assignments, truncating to 1501 for weight, removing 0
+// ii) DEBUG runtime::election-provider: 🗳 from 931 assignments, truncating to 755 for length, removing 176
+//
+// Thus, the only way to ensure that trimming actually works.
+async fn has_trimming_output(mut miner: KillChildOnDrop) -> bool {
```
niklasad1, line 52: I think we can remove this ugly stuff, if the solution is correct and accepted by the chain it's sufficient as we ensure that trimming always occurs on the staking-miner-playground

staking-miner-playground/runtime/build.rs
```diff
@@ -1,5 +1,9 @@
 use substrate_wasm_builder::WasmBuilder;
 
+pub const NOMINATORS: Option<&str> = option_env!("N");
+pub const CANDIDATES: Option<&str> = option_env!("C");
+pub const VALIDATORS: Option<&str> = option_env!("V");
```
jsdw, line 5: So, the N, C and V env vars can be set _at compile time_ to determine the value of these things, and we don't mind that we can't configure them at runtime?

niklasad1: yupp, that's the best option right now :(

staking-miner-playground/runtime/src/lib.rs
kianenigma: I recall configuring this with a goal like this: 

let's say you want 10 validators. provide 10 candidates, all of which will be winners. Given each of them a large number of evenly distributed voters, and then trim down. Technically this should not lead to an invalid solution. 

Ofc, you need to set the `MinerMaxWeight` such that we trim just a little bit, not too much. You can set that by feeding `MinerConfig::solution_weight()` repeatedly and finding a `v` for which we partially trim. 

https://github.com/paritytech/staking-miner-v2/blob/538/staking-miner-playground/runtime/src/lib.rs#L567

For example, If `MinerMaxweight` is 100, and you know `solution_weight(10, 100) == 120`, you want to find the biggest `x` for which `solution_weight(10, x) <= 100`.

staking-miner-playground/runtime/build.rs
kianenigma: interesting, are these constants then accessible at runtime for reading?

tests/monitor.rs
kianenigma: hmmm, not convinced. If you mis-configure this such that the trimming functions do nothing, how can we identify that.

kianenigma:
Looks mostly good, thanks!

tests/monitor.rs
niklasad1: yeah true, we need to have this and ensure it's not trimming output != 0 for weight and length

staking-miner-playground/runtime/build.rs
jsdw: Just to check; if Subxt _could_ read the constants from the node, would that help?

(I know it's not something we normally need to do, but I geuss I'm not feeling too opposed to the idea that when you read a constant, you can opt to load it from the node rather than the metadata; for instance you can make the client use older metadata or not subscribe to runtime updates or whatever, so this would be a good "fallback" option to provide anyway IMO)

staking-miner-playground/runtime/src/lib.rs
niklasad1: yeah, I copy-pasted it from substrate

staking-miner-playground/runtime/build.rs
jsdw: (update: as far as we can see, there's no way to read newer constants from a node, so alas :))

staking-miner-playground/runtime/src/lib.rs
```diff
@@ -450,15 +461,41 @@ parameter_types! {
...
-		.max_total
-		.unwrap();
-	pub MinerMaxLength: u32 = Perbill::from_rational(8u32, 10) *
-		*(<<Runtime as frame_system::Config>::BlockLength as Get<limits::BlockLength>>::get())
-		.max
-		.get(DispatchClass::Normal);
+
+	// This is a hack to get the number of validators, candidates and nominators
+	// used by node which uses the same env flags as the chain spec builder in the node crate.
+	Nominators: u32 = option_env!("N").unwrap_or("1000").parse().expect("env variable `N` must be number");
```
niklasad1, line 467: Nominators and Candidates are not used anymore.

Remove or should we configure something else in staking to work properly?

gpestana (changes requested):
Added a few comments and questions. Also I don't see to be able to build and run the tests for `staking-miner-playground` locally, the compilation is failing.

staking-miner-playground/runtime/src/lib.rs
niklasad1: nothing let's remove it

tests/monitor.rs
```diff
@@ -4,82 +4,91 @@
...
+
+		if now.elapsed() > MAX_DURATION_FOR_SUBMIT_SOLUTION {
 			break
 		}
 	}
 
-	assert!(success);
+	assert!(got_truncate_weight, "Trimming weight logs were not found");
+	assert!(got_truncate_len, "Trimming length logs were not found");
+
```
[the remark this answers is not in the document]
niklasad1: Those are not deterministic right now, I guess it's because how the nominators are generated with some randomness so the number of assignments differ between the runs.

src/epm.rs
```diff
@@ -207,6 +207,14 @@ where
 	let voters = snapshot.voters.clone();
 	let targets = snapshot.targets.clone();
 
+	log::trace!(
+		target: LOG_TARGET,
+		"mine solution: desired_target={}, voters={}, targets={}",
```
niklasad1, line 212:
```suggestion
		"mine solution: desired_targets={}, voters={}, targets={}",
```

niklasad1:
merging this now, because I want to run it our daily integration tests.

PR #539, opened 2023-04-27
Allow configuring winner count in dry-run

### LINKED ISSUE #461: add `simulate` command
The idea is to have a way to "simulate" the validator set with different configurations such size of the "winning set" and see how it affect the score and similar.

See https://polkadot.polkassembly.io/proposal/23

jsdw:
Allow passing winner-count into the `dry-run` command, and make providing the account ID optional.

Does anything else need adding/doing to make this useful (eg other config to allow overriding)? (I think in that case I might be tempted to have a separate `simulate` command like the original issue suggests)

Closes #461


lexnv (approved):

artifacts/metadata.scale
niklasad1, line 1: we need to revert this before merging :P

src/epm.rs
```diff
@@ -221,27 +226,70 @@ where
...
+			score,
+			solution_or_snapshot_size,
+		}),
 		Ok(Err(err)) => Err(Error::Other(format!("{:?}", err))),
 		Err(err) => Err(err.into()),
 	}
 }
 
+/// The result of calling [`fetch_snapshot_and_mine_solution`].
+pub struct MinedSolution<T: MinerConfig> {
```
niklasad1, line 244: nice :+1:

niklasad1 (approved):

src/commands/dry_run.rs
```diff
@@ -96,17 +97,34 @@ where
...
-
-	let outcome = api.rpc().dry_run(xt.encoded(), config.at).await?;
-
-	log::info!(target: LOG_TARGET, "dry-run outcome is {:?}", outcome);
-
-	match outcome {
-		Ok(()) => Ok(()),
-		Err(e) => Err(Error::Other(format!("{e:?}"))),
+	// Now we've logged the score, check whether the solution makes sense:
+	miner_solution.feasibility_check()?;
```
niklasad1, line 101: I realized that this fail when `--force-winner-count` is enabled because it would fetch `MinerConfig::MaxWinners` in check and always fails.

We can remove this completely or overwrite `MinerConfig::MaxWinners`

jsdw: Yup, it would!

I move it to after the logging for this reason. I figured that when doing a dry run, you might be interested to know about this failure either way (maybe you set `--force-winner-count` to the number of winners you expect the chain to be happy with or osmething).

I did also ponder having it in the `if let Some` block below; that way, we would only fail if we were otherwise already going to try and submit the thing for the node to check. But it felt like one might want to know this regardless

niklasad1: as we read the `MaxWinners` from the metadata and one would provide not use the same values in the metadata that will always fails.

so either just do the feasibility_check if `force_winner_count.is_some()` or add this in the beginning of this function:

```rust
	if Some(max_winners) = config.force_winner_count {
		static_types::MaxWinners.set(max_winners);
	}
```

jsdw: Okeydoke; I disabled if `.is_none()`!

PR #540, opened 2023-04-27
`commands` module to expose each sub-command, and put options next to them

jsdw:
This is a slightly opinionated reshuffle of things to:
- Make it clear what the main entry points (subcommands) are by putting them into a `commands` module.
- Put the options for each subcommand next to it (thinning out the `opts` module somewhat; maybe we can go fruther because not everything in that is related to CLI opts by the looks of it).
- Remove a couple of things from `prelude` (just because personally I think it's easier to write but harder to read/follow where things come from when `prelude::*` is used for too many imports).

What do you reckon?

### DIFF
--- a/src/commands/dry_run.rs
+++ b/src/commands/dry_run.rs
@@ -18,8 +18,35 @@
 
 use pallet_election_provider_multi_phase::RawSolution;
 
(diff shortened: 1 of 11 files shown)


niklasad1 (approved):
Looks good to me

gpestana (approved):

PR #557, opened 2023-05-07
Implements emergency solution command

gpestana:
Implements the `emergency_solution_cmd`, similar to the staking-miner-v1 implementation. In addition, it saves the encoded `set_emergency_solution` EPM call into a file for easy handling.

### DIFF
--- a/src/commands/emergency_solution.rs
+++ b/src/commands/emergency_solution.rs
@@ -19,6 +19,7 @@
 use crate::{epm, error::Error, opt::Solver, prelude::*, static_types};
 use clap::Parser;
 use codec::Encode;
+use sp_core::hexdisplay::HexDisplay;
 use std::io::Write;
 use subxt::tx::TxPayload;
 
@@ -88,7 +89,6 @@ where
 
 	let call = epm::set_emergency_result(supports.clone())?;
 	let encoded_call = call.encode_call_data(&api.metadata())?;
-
 	let encoded_supports = supports.encode();
 
 	// write results to files.
@@ -97,6 +97,10 @@ where
 	supports_file.write_all(&encoded_supports)?;
 	encoded_call_file.write_all(&encoded_call)?;
 
+	let hex = HexDisplay::from(&encoded_call);
+	log::info!(target: LOG_TARGET, "Hex call:\n {:?}", hex);
+
+	log::info!(target: LOG_TARGET, "Use the hex encoded call above to construct the governance proposal or the extrinsic to submit.");
 	log::info!(target: LOG_TARGET, "ReadySolution: size {:?} / score = {:?}", encoded_size, score);
 	log::info!(target: LOG_TARGET, "`set_emergency_result` encoded call written to ./encoded.call");
 
--- a/src/commands/emergency_solution.rs
+++ b/src/commands/emergency_solution.rs
@@ -16,8 +16,10 @@
 
 //! The emergency-solution command.
 
-use crate::{error::Error, opt::Solver, prelude::*, static_types};
+use crate::{epm, error::Error, opt::Solver, prelude::*, static_types};
 use clap::Parser;
+use codec::Encode;
+use std::io::Write;
 
 #[derive(Debug, Clone, Parser)]
 #[cfg_attr(test, derive(PartialEq))]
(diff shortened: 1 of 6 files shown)


src/commands/emergency_solution.rs
```diff
@@ -45,5 +47,51 @@ where
...
+			force_winner_count
+		);
+		supports.sort_unstable_by_key(|(_, s)| s.total);
+		supports.truncate(force_winner_count as usize);
+	}
+
+	// write to file and stdout.
+	let encoded_supports = supports.encode();
+	let mut supports_file = std::fs::File::create("solution.supports.bin")?;
+	supports_file.write_all(&encoded_supports)?;
```
kianenigma, line 87: Let's print the entire call ready for submission as well?

kianenigma (approved):

src/commands/emergency_solution.rs
```diff
@@ -45,5 +47,51 @@ where
...
+
+	log::info!(target: LOG_TARGET, "ReadySolution: size {:?} / score = {:?}", encoded_size, score);
+	log::trace!(
+		target: LOG_TARGET,
+		"Supports: {}",
```
Ank4n, line 92: Current code would [truncate](https://github.com/paritytech/substrate/blob/master/primitives/core/src/hexdisplay.rs#L40) the hex display. Should we use debug display? 

I also made this [small PR](https://github.com/paritytech/polkadot/pull/7194/files) for miner v1 printing encoded call instead of `supports`.

```suggestion
		"Supports: {:?}",
```

Ank4n: Might be better to save it in a file as well.

src/commands/emergency_solution.rs
```diff
@@ -45,5 +47,51 @@ where
...
+	let miner_solution = epm::fetch_snapshot_and_mine_solution::<T>(
+		&api,
+		config.at,
+		config.solver,
+		round,
+		config.force_winner_count,
+	)
+	.await?;
+
+	let ready_solution = miner_solution.feasibility_check()?;
```
niklasad1, line 66: Because `MaxWinners` is read from the metadata and if one enables `--force-winner-count <val>` this might differ from the metadata and feasiblity_check reads the value from the metadata.

so either just do the feasibility_check if `force_winner_count.is_none()` or add the following in the beginning of this function:

```rust
	if Some(max_winners) = config.force_winner_count {
		static_types::MaxWinners::set(config.force_winner_count);
	}
```

niklasad1 (commented):
looks good,

just one thing regarding metadata on `MaxWinners`

src/epm.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) async fn update_metadata_constants(api: &SubxtClient) -> Result<(), E
 	const SIGNED_MAX_WEIGHT: EpmConstant = EpmConstant::new("SignedMaxWeight");
 	const MAX_LENGTH: EpmConstant = EpmConstant::new("MinerMaxLength");
 	const MAX_VOTES_PER_VOTER: EpmConstant = EpmConstant::new("MinerMaxVotesPerVoter");
-	const MAX_WINNERS: EpmConstant = EpmConstant::new("MinerMaxWinners");
+	const MAX_WINNERS: EpmConstant = EpmConstant::new("MaxWinners");
```
niklasad1, line 70: Ok, this was set to work with older metadata right?

Smart move as this is just duplicated in the `MinerConfig` :P

niklasad1: Maybe add a comment for that :)

niklasad1 (approved):

Ank4n (approved):

src/epm.rs
```diff
@@ -112,6 +112,16 @@ fn read_constant<'a, T: serde::Deserialize<'a>>(
 	})
 }
 
+/// Helper to construct a set emergency solution transaction.
+pub(crate) fn set_emergency_result<A: Encode + TypeInfo + 'static>(
+	supports: frame_election_provider_support::Supports<A>,
+) -> Result<DynamicTxPayload<'static>, Error> {
+	let scale_result = to_scale_value(supports)
+		.map_err(|e| Error::DynamicTransaction(format!("Failed to decode `Supports`: {:?}", e)))?;
```
niklasad1, line 120:
```suggestion
		.map_err(|e| Error::DynamicTransaction(format!("Failed to encode `Supports`: {:?}", e)))?;
```

src/epm.rs
niklasad1, line 70:
```suggestion
    // NOTE: `MaxWinners`  is used instead of `MinerMaxWinners` to work with older metadata
  	const MAX_WINNERS: EpmConstant = EpmConstant::new("MaxWinners");
```

src/epm.rs
```diff
@@ -67,7 +67,8 @@ pub(crate) async fn update_metadata_constants(api: &SubxtClient) -> Result<(), E
 	const SIGNED_MAX_WEIGHT: EpmConstant = EpmConstant::new("SignedMaxWeight");
 	const MAX_LENGTH: EpmConstant = EpmConstant::new("MinerMaxLength");
 	const MAX_VOTES_PER_VOTER: EpmConstant = EpmConstant::new("MinerMaxVotesPerVoter");
-	const MAX_WINNERS: EpmConstant = EpmConstant::new("MinerMaxWinners");
+    // NOTE: `MaxWinners`  is used instead of `MinerMaxWinners` to work with older metadata
```
niklasad1, line 70:
```suggestion
    // NOTE: `MaxWinners` is used instead of `MinerMaxWinners` to work with older metadata
```

PR #571, opened 2023-05-19
update subxt

niklasad1:


src/prelude.rs
```diff
@@ -65,32 +65,35 @@ pub type Config = subxt::PolkadotConfig;
...
-		type = "pallet_election_provider_multi_phase::Phase",
-		with = "::pallet_election_provider_multi_phase::Phase"
+		path = "pallet_election_provider_multi_phase::Phase",
+		with = "::subxt::utils::Static<::pallet_election_provider_multi_phase::Phase<u32>>"
 	),
 	substitute_type(
-		type = "pallet_election_provider_multi_phase::SolutionOrSnapshotSize",
-		with = "::pallet_election_provider_multi_phase::SolutionOrSnapshotSize"
+		path = "pallet_election_provider_multi_phase::RawSolution<S>",
+		with = "::subxt::utils::Static<::pallet_election_provider_multi_phase::RawSolution<S>>"
```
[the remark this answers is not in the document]
jsdw: So I ran codegen to generate the code and pasted it in place of your macro usage. This make it a lot clearer what lines in the generated code are causing an issue:

```shell
subxt% cargo run -- codegen --file ../staking-miner-v2/artifacts/metadata.scale \
  --substitute-type='sp_arithmetic::per_things::PerU16=::subxt::utils::Static<::subxt::ext::sp_runtime::PerU16>' \
  --substitute-type='sp_npos_elections::ElectionScore=::subxt::utils::Static<::sp_npos_elections::ElectionScore>' \
  --substitute-type='pallet_election_provider_multi_phase::SolutionOrSnapshotSize=::subxt::utils::Static<::pallet_election_provider_multi_phase::SolutionOrSnapshotSize>' \
  --substitute-type='pallet_election_provider_multi_phase::Phase=::subxt::utils::Static<::pallet_election_provider_multi_phase::Phase<u32>>' \
  --substitute-type='pallet_election_provider_multi_phase::RawSolution<S>=::subxt::utils::Static<::pallet_election_provider_multi_phase::RawSolution<S>>' \
  | rustfmt > OUTPUT.rs

```

So the first error that pops up is to do with this struct:

```rust
use super::runtime_types;
#[derive(
	:: subxt :: ext :: codec :: Decode,
	:: subxt :: ext :: codec :: Encode,
	:: subxt :: ext :: scale_decode :: DecodeAsType,
	:: subxt :: ext :: scale_encode :: EncodeAsType,
	Debug,
)]
# [codec (crate = :: subxt :: ext :: codec)]
#[decode_as_type(crate_path = ":: subxt :: ext :: scale_decode")]
#[encode_as_type(crate_path = ":: subxt :: ext :: scale_encode")]
pub struct SignedSubmission<_0, _1, _2> {
	pub who: _0,
	pub deposit: _1,
	pub raw_solution: ::subxt::utils::Static<
		::pallet_election_provider_multi_phase::RawSolution<_2>,
	>,
	pub call_fee: _1,
	#[codec(skip)]
	pub __subxt_unused_type_params: ::core::marker::PhantomData<_2>,
}
```

Somehow `parity-scale-codec` knows to assume that `_0` and `_1` must be bounded by `Encode` and `Decode`, but not `_2`. Adding the manual bounds for it fixes it for me. Obviously not very helpful though since there's no way to do that using the macro. Potentially a tweak to `parity-scale-codec` could resolve this.

Otherwise, I think if you want to substitute `::pallet_election_provider_multi_phase::RawSolution`, you'd have to forego using `Static` here and substitute it with a custom `RawSolution` type which implements `EncodeAsType` and `DecodeAsType` like:

```rust
#[derive(Encode, Decode, EncodeAsType, DecodeAsType, Debug)]
pub struct RawSolution<S> {
	pub solution: subxt::utils::Static<S>,
	pub score: subxt::utils::Static<ElectionScore>,
	pub round: u32
}
```

(The `Static`s here aren't ideal, and maybe this would hit some other error, but it's the simplest thing I can think of suggesting trying offhand. I note that `S` is always `NposCompactSolution16`).

Now, the next issue:

```rust
#[derive(
	:: subxt :: ext :: codec :: Decode,
	:: subxt :: ext :: codec :: Encode,
	:: subxt :: ext :: scale_decode :: DecodeAsType,
	:: subxt :: ext :: scale_encode :: EncodeAsType,
	Debug,
)]
# [codec (crate = :: subxt :: ext :: codec)]
#[decode_as_type(crate_path = ":: subxt :: ext :: scale_decode")]
#[encode_as_type(crate_path = ":: subxt :: ext :: scale_encode")]
pub struct NposCompactSolution16 {
	pub votes1: ::std::vec::Vec<(
		::subxt::ext::codec::Compact<::core::primitive::u32>,
		::subxt::ext::codec::Compact<::core::primitive::u16>,
	)>,
	pub votes2: ::std::vec::Vec<(
		::subxt::ext::codec::Compact<::core::primitive::u32>,
		(
			::subxt::ext::codec::Compact<::core::primitive::u16>,
			::subxt::ext::codec::Compact<
				::subxt::utils::Static<::subxt::ext::sp_runtime::PerU16>,
			>,
		),
		::subxt::ext::codec::Compact<::core::primitive::u16>,
	)>,
	// ...
}
```

Here, the error message/location is less clear, but I suspect the issue is the:

```rust
::subxt::ext::codec::Compact<
	::subxt::utils::Static<::subxt::ext::sp_runtime::PerU16>,
>,
```

Ie the `Static` inside the `Compact` probably confuses things. I can see about allowing `Static` to impl `CompactAs` (and whatever else needed) so long as the thing inside does, which I hope would allow it to appear inside `Compact` and be fairly transparent.

A potential solution is just writing a manual version of this `NposCompactSolution16` type and swapping the whole thing, but yeah not ideal.

I guess the easiest solution though given these issues is to either:

1. not substitute these troublesome types at all, and write `From` impls or whatever to make converting between the versions you have, and the versions that are generated, easy to do. This gives the best type safety and such (each use of `Static` degrades Subxt's ability to adapt types as needed for the node it's talking to).
2. use dynamic types instead; you forego type checking completely in this case, but potentially you can be even more flexible in your types to accomodate for runtime differences or whatever.

src/commands/monitor.rs
```diff
@@ -511,10 +514,14 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
 		.create_signed_with_nonce(&tx, &*signer, nonce, ExtrinsicParams::default())?;
 
 	if dry_run {
-		match api.rpc().dry_run(xt.encoded(), None).await? {
-			Ok(()) => (),
-			Err(e) => return Err(Error::TransactionRejected(format!("{e:?}"))),
-		};
+		let dry_run_bytes = api.rpc().dry_run(xt.encoded(), Some(block_hash)).await?;
+
+		match dry_run_bytes.into_dry_run_result(&api.metadata())? {
```
niklasad1, line 519: /cc @jsdw I reckon that this is expensive and I didn't find a better way, how slow is it to read the RwLock and clone the metadata?

src/commands/emergency_solution.rs
```diff
@@ -26,7 +26,7 @@ use subxt::tx::TxPayload;
 #[derive(Debug, Clone, Parser)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct EmergencySolutionConfig {
-	/// The block hash at which scraping happens. If none is provided, the latest head is used.
+	/// The block hash at which scraping happens
```
niklasad1, line 29:
```suggestion
	/// The block hash at which scraping happens. If none is provided, the latest head is used.
```

src/commands/monitor.rs
jsdw: Ooh I forgot about that!  Yeah, I wouldn't worry about it; this is no less efficient than any other approach :)

jsdw (approved):

Cargo.toml
```diff
@@ -37,7 +37,7 @@ once_cell = "1.17"
 [dev-dependencies]
 anyhow = "1"
 assert_cmd = "2.0"
-sp-storage = "11.0.0"
+sp-storage = { git = "https://github.com/paritytech/substrate" }
```
lexnv, line 40: nit: Would the latest release `13.0.0` work here?

lexnv (approved):

Cargo.toml
niklasad1: yeah but I changed it because it doesn't make sense to make the dependency tree even bigger :)

niklasad1: I need to release a bunch crates for that and my goal is to use `master branches` here and check out crates.io version on each release.

Because it's so annoying to patch crates if we need some specific from substrate that isn't released.

PR #575, opened 2023-05-24
chore: remove unused deps and features

niklasad1:


Cargo.toml
```diff
@@ -23,13 +23,10 @@ subxt = { git = "https://github.com/paritytech/subxt" }
 scale-value = "0.6.0"
 
 # substrate
-sp-io = { git = "https://github.com/paritytech/substrate" }
-frame-system = { git = "https://github.com/paritytech/substrate" }
```
lexnv, line 27: nit: Maybe we could transform unused crates into a compile-time error with `#![deny(unused_crate_dependencies)]`, that may be a bit restrictive while experimenting with a new feature, leaving this up to you 🙏

lexnv (approved):
LGTM 👍

Cargo.toml
niklasad1: good comment but it doesn't work well for `dev-dependencies` I could add `cargo machete instead` :)

PR #587, opened 2023-06-01
improve README

niklasad1:
Close https://github.com/paritytech/staking-miner-v2/issues/451

PR #588, opened 2023-06-01
chore: release v1.0.0

niklasad1:
Blocked on #587

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -246,7 +246,7 @@ async fn runtime_upgrade_task(api: SubxtClient, tx: oneshot::Sender<Error>) {
 				log::info!(target: LOG_TARGET, "upgrade to version: {} successful", version);
 			},
 			Err(e) => {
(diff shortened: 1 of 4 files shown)


lexnv (approved):

gpestana (approved):

src/main.rs
```diff
@@ -246,7 +246,7 @@ async fn runtime_upgrade_task(api: SubxtClient, tx: oneshot::Sender<Error>) {
 				log::info!(target: LOG_TARGET, "upgrade to version: {} successful", version);
 			},
 			Err(e) => {
-				log::warn!(target: LOG_TARGET, "upgrade to version: {} failed: {:?}", version, e);
+				log::debug!(target: LOG_TARGET, "upgrade to version: {} failed: {:?}", version, e);
```
niklasad1, line 249: This will only fail if the applied update is the same as the current version.

Because substrate emit one at startup this will occur on every start so downgrading it debug.

PR #589, opened 2023-06-02
change `submission strategy == if leading` to not submit equal score

niklasad1:
This code was copied from the staking-miner v1 but I think `submission strategy == if leading`  should really only submit if the score is strictly better than existing scores on chain. 

Thus, this is confusing and not a good default strategy such as if there is already a solution with the same score one might lose transaction fee when submitting then.

This boils down to whether one trusts that previous scores are legit or not but as we already have `submission strategy == always` that can be used if one doesn't trust the scores on chain.

//cc @kianenigma do you remember why we did it like that?

### DIFF
--- a/src/commands/monitor.rs
+++ b/src/commands/monitor.rs
@@ -609,8 +609,7 @@ pub(crate) fn score_passes_strategy(
 	match strategy {
 		SubmissionStrategy::Always => true,
 		SubmissionStrategy::IfLeading =>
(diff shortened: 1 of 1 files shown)


niklasad1:
FYI, I was contacted by an existing user that was surprised by this as well.

kianenigma (approved):
Yeah I agree, confusing.

PR #687, opened 2023-09-20
refactor(trimming): implement pre-trimming

niklasad1:
This commit refactors trimming such that the staking-miner is responsible to ensure that trimming doesn't have to be done be on-chain.

The pre-trimming is implemented as follows:

1. Weight trimming: Initially the voters are iteratively removed until Weight < MaxWeight by using MinerConfig::solution_weight for this part a "new solution" doesn't need to be "mined" and is much more efficient than part 2 and only has to be performed once.
2. Length Trimming: Instead of removing one voter at the time and then “mine a new solution” as the initial implementation did. Now the implementation removes the number of voters in a binary search manner to find "the biggest set of voters" where no trimming occurs such that it doesn't get trimmed.

By “binary search manner” I mean:

```rust
let  mut  l  =  1;
let  mut  h  =  voters.len();
let  mut  best_solution  =  None;

while  l  <=  h {
   let  mid  = ((h  -  l) /  2) +  l;
   let next_state = voters.trim(mid)?;
   let solution = mine_solution(next_state);
	 
   if !solution.is_trimmed() {
     best_solution  =  Some((solution, score, solution_or_snapshot_size));
     h  =  mid  -  1;
   } else {
     l  =  mid  +  1;
   }
}
```

This makes things much faster than the iterative approach I implemented first

This PR also adds a prometheous metric to count the number of "pre-trimming length ops"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,11 +24,11 @@ scale-value = "0.10.0"
 
 
 # substrate
-frame-election-provider-support = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
-pallet-election-provider-multi-phase = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
-sp-npos-elections = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
-frame-support = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
-sp-runtime = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
+frame-election-provider-support = { git = "https://github.com/paritytech/polkadot-sdk" , branch = "na-epm-trimming-api" }
+pallet-election-provider-multi-phase = { git = "https://github.com/paritytech/polkadot-sdk" , branch = "na-epm-trimming-api" }
+sp-npos-elections = { git = "https://github.com/paritytech/polkadot-sdk/" , branch = "na-epm-trimming-api" }
+frame-support = { git = "https://github.com/paritytech/polkadot-sdk" , branch = "na-epm-trimming-api" }
+sp-runtime = { git = "https://github.com/paritytech/polkadot-sdk" , branch = "na-epm-trimming-api" }
 
 # prometheus
 prometheus = "0.13"
@@ -38,7 +38,7 @@ once_cell = "1.18"
 [dev-dependencies]
 anyhow = "1"
 assert_cmd = "2.0"
-sp-storage = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
+sp-storage = { git = "https://github.com/paritytech/polkadot-sdk" , branch = "na-epm-trimming-api" }
 regex = "1"
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,12 +22,13 @@ pin-project-lite = "0.2"
 subxt = "0.29"
 scale-value = "0.10.0"
 
+
 # substrate
-frame-election-provider-support = { git = "https://github.com/paritytech/substrate" }
-pallet-election-provider-multi-phase = { git = "https://github.com/paritytech/substrate" }
-sp-npos-elections = { git = "https://github.com/paritytech/substrate" }
-frame-support = { git = "https://github.com/paritytech/substrate" }
-sp-runtime = { git = "https://github.com/paritytech/substrate" }
+frame-election-provider-support = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
+pallet-election-provider-multi-phase = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
+sp-npos-elections = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
+frame-support = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
+sp-runtime = { git = "https://github.com/paritytech/substrate" , branch = "na-epm-trimming-api" }
 
 # prometheus
 prometheus = "0.13"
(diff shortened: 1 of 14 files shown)


staking-miner-playground/runtime/src/lib.rs
```diff
@@ -343,7 +332,7 @@ impl<const PERIOD: BlockNumber> ShouldEndSession<BlockNumber>
 	}
 }
 
-const SESSION: BlockNumber = 1 * MINUTES;
+const SESSION: BlockNumber = 3 * MINUTES;
```
niklasad1, line 335: Because the trimming is slower with "pre-trimming", I increased the SignPhase a bit...

staking-miner-playground/runtime/src/lib.rs
```diff
@@ -465,32 +454,12 @@ parameter_types! {
...
-		Perbill::from_percent(90) * BlockLength::get()
-	);
+	// NOTE: This value must be the same as `V` in `node/src/chainspec.rs`.
+	Validators: u32 = option_env!("V").unwrap_or("20").parse().expect("env variable `V` must be number");
 
-	pub MinerMaxWeight: Weight = prod_or_enforce_trimming!(
-		Perbill::from_rational(8u32, 10) * <Runtime as frame_system::Config>::BlockWeights::get().get(DispatchClass::Normal).max_total.unwrap(),
-		WeightTrimming::get()
-	);
+	pub MinerMaxLength: u32 = Perbill::from_percent(60) * *(<<Runtime as frame_system::Config>::BlockLength as Get<limits::BlockLength>>::get()).max.get(DispatchClass::Normal);
```
niklasad1, line 461: modified these values to ensure trimming always occurs when we run it.

Cargo.toml
```diff
@@ -22,12 +22,13 @@ pin-project-lite = "0.2"
...
 scale-value = "0.10.0"
 
+
 # substrate
-frame-election-provider-support = { git = "https://github.com/paritytech/substrate" }
-pallet-election-provider-multi-phase = { git = "https://github.com/paritytech/substrate" }
-sp-npos-elections = { git = "https://github.com/paritytech/substrate" }
-frame-support = { git = "https://github.com/paritytech/substrate" }
-sp-runtime = { git = "https://github.com/paritytech/substrate" }
+frame-election-provider-support = { git = "https://github.com/paritytech/polkadot-sdk" , branch = "na-epm-trimming-api" }
```
niklasad1, line 27: change to master once we have merged the substrate PR

src/epm.rs
```diff
@@ -62,6 +66,54 @@ impl std::fmt::Display for EpmConstant {
 	}
 }
 
+/// Represent voters that may be trimmed
+///
+/// The trimming works by removing the voter with the least amount of stake
```
gpestana, line 71:
```suggestion
/// The trimming works by removing the voter with the least amount of stake.
```

src/epm.rs
```diff
@@ -62,6 +66,54 @@ impl std::fmt::Display for EpmConstant {
...
+		let mut voters_by_stake = BTreeMap::new();
+
+		for (idx, (_voter, stake, _supports)) in voters.iter().enumerate() {
+			voters_by_stake.insert(*stake, idx);
+		}
+
+		Self { voters, voters_by_stake }
+	}
+
+	/// Trim the next voter with the least amount of stake.
```
gpestana, line 94: Have we tried to trim more than one voter in each iteration? This may reduce the number of iterations required to finalise the trimming. Maybe we could have an heuristic where depending on how much the solution exceeds the limit, trim 1 or more voters in one go.

This probably should not block this PR, but perhaps addressed in a subsequent PR if helpful.

src/epm.rs
```diff
@@ -219,47 +271,62 @@ where
...
+			},
+			Ok(Ok((solution, _score, _, _))) => {
+				if solution.unique_targets().len() != desired_targets as usize {
+					return Err(Error::Feasibility(format!(
+						"Invalid winner count {}, expected {desired_targets}",
+						solution.unique_targets().len()
+					)))
+				}
+
+				voters.trim_next()?;
```
gpestana, line 325: As per my comment above, perhaps we could pass the number of voters to trim which will depend on the iteration we are at. E.g. if this is the first call to `fn trim_next` in the loop, trim N voters. After a few attempts, start to trim only one voter since we're approaching the solution that does not need any more trimming.

src/epm.rs
gpestana: We could also calculate the difference between the `weight_trimmed`/`length_trimmed` calculated in  `prepare_election_result_with_snapshot` and use that as an heuristic of how much trimming will be required. Then we can use that to request the trimming of N voters in one go, if there's still a long way to trim.

gpestana (approved):
Added a few comments that could optimise the number of trimming round before reaching the appropriate solution weight/length, but I think that should not be a blocker for this PR if it's too much work.

Thanks for implementing this!

src/epm.rs
niklasad1: That's good point, I think we should merge this PR and create a issue for your idea.

If there's much trimming to do then yeah this would be very slow as it does it one-by-one.

src/epm.rs
```diff
@@ -62,6 +72,112 @@ impl std::fmt::Display for EpmConstant {
...
+			// Remove votes for an account.
+			for (_voter, _stake, supports) in &mut voters {
+				supports.retain(|a| a != &rm);
+				targets.extend(supports);
+			}
+
+			let desired_targets = desired_targets;
+			let targets = targets.len() as u32;
+
+			let est_weight: Weight = tokio::task::spawn_blocking(move || {
```
niklasad1, line 136: decided to add this in the constructor/new because it should only be needed to do once

src/epm.rs
```diff
@@ -219,47 +371,69 @@ where
...
+		})
+	}
+
+	prometheus::on_trim_attempt();
+
+	let mut l = 1;
+	let mut h = voters.len();
+	let mut best_solution = None;
+
+	while l <= h {
```
niklasad1, line 402: binary search to find the best "pre-trim"

PR #715, opened 2023-11-22
chore: update deps to publish on crates.io

niklasad1:
Preparation for a new release which mainly updates subxt and jsonrpsee.

One nice thing is that default port number in the URI can be omitted now


src/main.rs
```diff
@@ -135,35 +137,43 @@ async fn main() -> Result<(), Error> {
...
 		let fut = match command {
-			Command::Monitor(cfg) => commands::monitor_cmd::<MinerConfig>(api, cfg).boxed(),
-			Command::DryRun(cfg) => commands::dry_run_cmd::<MinerConfig>(api, cfg).boxed(),
+			Command::Monitor(cfg) => commands::monitor_cmd::<MinerConfig>(client, cfg).boxed(),
+			Command::DryRun(cfg) => commands::dry_run_cmd::<MinerConfig>(client, cfg).boxed(),
 			Command::EmergencySolution(cfg) =>
-				commands::emergency_solution_cmd::<MinerConfig>(api, cfg).boxed(),
+				commands::emergency_solution_cmd::<MinerConfig>(client, cfg).boxed(),
 			Command::Info => async {
-				let is_compat = if runtime::validate_codegen(&api).is_ok() { "YES" } else { "NO" };
```
niklasad1, line 160: /cc @jsdw how to do this now? 

I could not find this API anymore?!

artifacts/metadata.scale
niklasad1, line 1: I was forced to update the metadata for the codegen to work

lexnv: Hmm interesting, might be worth a subxt issue probably we'd want to keep backwards compatibility between 14 and 15 versions 🤔

lexnv (approved):

artifacts/metadata.scale
niklasad1: I decided not to trim the metadata because of https://github.com/paritytech/subxt/issues/1286

src/client.rs
```diff
@@ -0,0 +1,67 @@
+use crate::prelude::*;
+use jsonrpsee::ws_client::WsClientBuilder;
+use subxt::backend::rpc::RpcClient as RawRpcClient;
+
+/// Wraps the subxt interfaces to make it easy to use for the staking-miner.
+#[derive(Clone, Debug)]
```
niklasad1, line 6: subxt's APIs has been modified a bit, thus this type is introduced to provide a wrapper for the APIs that the staking miner needs.

gpestana (approved):

PR #720, opened 2023-11-28
use `subxt-signer` to reduce the number of deps

### LINKED ISSUE #718: deps: remove `subxt --feature substrate-compat` to remove duplicated dependencies
Currently, staking-miner both have its own polkadot-sdk dependencies and polkadot-sdk dependencies via subxt which end up with "same dependency" but with different versions which would be neat to get rid of.

niklasad1:
Close #718

This commit reduces the dependency tree ~50 crates by fixing that the repo doesn't bring in two versions of sp-core and sp-runtime.

The reason is that subxt has its own version of these crates when the `substrate-compat` feature flag is enabled and ideally we would remove both `sp-core` and `sp-runtime` completely but it doesn't work because `pallet-epm` depends on them.


README.md
```diff
@@ -1,33 +1,36 @@
...
-This is a re-write of the [polkadot staking miner](https://github.com/paritytech/polkadot/tree/master/utils/staking-miner) using [subxt](https://github.com/paritytech/subxt) to avoid hard dependency to each runtime version.
+This is a re-write of the [staking miner](https://github.com/paritytech/polkadot/tree/master/utils/staking-miner) using [subxt](https://github.com/paritytech/subxt) to avoid hard dependency to each runtime version.
 
 The binary itself embeds [static metadata](./artifacts/metadata.scale) to
 generate a rust codegen at compile-time that [subxt provides](https://github.com/paritytech/subxt).
 
-Runtime upgrades are handled by staking-miner-v2 by upgrading storage constants
-and that will work unless there is a breaking change in any of pallets used by
-the staking-miner (mainly pallet-election-provider and pallet-system are used).
+Runtime upgrades are handled by polkadot-staking-miner by upgrading storage constants
```
niklasad1, line 10:
```suggestion
Runtime upgrades are handled by the polkadot-staking-miner by upgrading storage constants
```

README.md
```diff
@@ -114,7 +136,7 @@ To update the metadata you need to connect to a polkadot, kusama or westend node
 $ cargo install --locked subxt-cli
 # Download the metadata from a local node and replace the current metadata
 # See `https://github.com/paritytech/subxt/tree/master/cli` for further documentation of the `subxt-cli` tool.
-$ subxt metadata -f bytes > artifacts/metadata.scale
+$ subxt metadata --url wss://rpc.polkadot.io --pallets "ElectionProviderMultiPhase,System" -f bytes > artifacts/metadata.scale
```
niklasad1, line 139:
```suggestion
$ subxt metadata --pallets "ElectionProviderMultiPhase,System" -f bytes > artifacts/metadata.scale
```

PR #726, opened 2023-12-12
chore(deps): bump subxt, subxt-signer, scale-value

niklasad1:


src/commands/monitor.rs
```diff
@@ -549,7 +551,7 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
 
 	match listen {
 		Listen::Head => {
-			let in_block = tx_progress.wait_for_in_block().await?;
+			let in_block = wait_for_in_block(tx_progress).await?;
```
niklasad1, line 554: This is no longer provided by subxt

.gitlab-ci.yml
```diff
@@ -15,7 +15,7 @@ default:
       - api_failure
 
 variables:
-  CI_IMAGE:                        "paritytech/ci-linux:production"
+  CI_IMAGE:                        "paritytech/ci-unified:bullseye-1.74.0"
```
alvicsam, line 18:
```suggestion
  CI_IMAGE:                        "`paritytech/ci-unified:bullseye-1.74.0-2023-11-01-v20231204`"
```
The latest one is currently unstable. If the image here should be the same as in polkadot-sdk we can add a snippet so they use the same image.

alvicsam (approved):

PR #728, opened 2023-12-13
refactor: make solution extrinsic mortal

### LINKED ISSUE #727: Consider adding reasonable mortality to `submit` extrinsic
Consider configure a reasonable mortality parameters in the `submit` extrinsic.

Related to https://github.com/paritytech-secops/srlabs_findings/issues/329

niklasad1:
Close #727

### DIFF
--- a/src/commands/monitor.rs
+++ b/src/commands/monitor.rs
@@ -527,6 +527,11 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
 	at: &Header,
 ) -> Result<(), Error> {
 	let tx = epm::signed_solution(RawSolution { solution, score, round })?;
+
+	// TODO: https://github.com/paritytech/polkadot-staking-miner/issues/730
+	//
+	// The extrinsic mortality length is static and doesn't know when the
+	// signed phase ends.
 	let signed_phase_len = client
 		.chain_api()
 		.constants()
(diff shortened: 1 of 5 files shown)


src/commands/monitor.rs
```diff
@@ -523,15 +524,22 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
...
 ) -> Result<(), Error> {
 	let tx = epm::signed_solution(RawSolution { solution, score, round })?;
-
-	let xt = client.chain_api().tx().create_signed_with_nonce(
-		&tx,
-		&signer,
-		nonce as u64,
-		Default::default(),
-	)?;
+	let signed_phase_len = client
```
niklasad1, line 530: NOTE: this is static applies the extrinsic to live for "signed_phase_len" and if for example the last block in the sign phase it will "live" for additional blocks.

We could change that as well but that is a slightly bigger change.

gpestana: Yes, I think this approximation is OK given that we also have the checks on the EPM side. 

otoh, maybe it would be easier to know exactly the boundaries of the signed phase in on terms of block number if the `Phase::Signed` in EPM wrapped the block number when it started? That way the miner could fetch the current EPM phase and know `mortal(signed_phase_started, signed_phase_len)`. is this worth the change?

gpestana (approved):
Added a comment regarding the approximation with number of blocks span of the tx and whether we could/should improve. Otherwise LGMT

src/commands/monitor.rs
niklasad1: Yeah, indeed it requires some changes because each block is executed in its own async task and needs synchronization.

Lemme create another issue for it :)

gpestana: Thanks! I think for now this and the EPM-side changes are enough to tackle the original issue.

lexnv (approved):

Issue #746, opened 2023-12-25
tagging docker images with release tag instead of commit hash


arminteimouri:
For `polkadot-staking-miner`, docker images are pushed with this naming convention: `branchName`-`7 characters of the commit SHA` <br>Exp:[main-b250ced](https://hub.docker.com/layers/paritytech/polkadot-staking-miner/main-b250ced/images/sha256-f2a98e7590446f4ee4e1621b92d0bf1ffe37cd333060fa46354782f43a182aa0?context=explore)<br>
As a result of [this line in the Job](https://github.com/paritytech/polkadot-staking-miner/blob/main/.github/workflows/ci.yml#L236). <br>
If we change the follwing, I think it will fix the issue.
```
echo IMAGE_TAG=$(if [ -n "$GITHUB_REF_NAME" ]; then echo "$GITHUB_REF_NAME"; else echo "main-${GITHUB_SHA::7}"; fi) >> $GITHUB_ENV
```

arminteimouri:
[PR](https://github.com/paritytech/polkadot-staking-miner/pull/747)

PR #747, opened 2023-12-25
changing image tag GITHUB_ENV

arminteimouri:
issue https://github.com/paritytech/polkadot-staking-miner/issues/746

> GITHUB_REF_NAME	The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, feature-branch-1.

> For pull requests, the format is refs/pull/<pr_number>/merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
 
       - name: Prepare Docker environment
         run: |
-          echo IMAGE_TAG=$(if [ "$GITHUB_REF" == "refs/heads/main" ]; then echo "main-${GITHUB_SHA::7}"; else echo "$GITHUB_REF_NAME"; fi) >> $GITHUB_ENV
+          echo IMAGE_TAG=$(if [ -n "$GITHUB_REF_NAME" ]; then echo "$GITHUB_REF_NAME"; else echo "main-${GITHUB_SHA::7}"; fi) >> $GITHUB_ENV
           echo PUSH_IMAGE=true >> $GITHUB_ENV
           echo "Docker image will be published with the tag: ${{ env.IMAGE_TAG }}!"
           chmod +x ./artifacts/polkadot-staking-miner


.github/workflows/ci.yml
```diff
@@ -233,7 +233,7 @@ jobs:
 
       - name: Prepare Docker environment
         run: |
-          echo IMAGE_TAG=$(if [ "$GITHUB_REF" == "refs/heads/main" ]; then echo "main-${GITHUB_SHA::7}"; else echo "$GITHUB_REF_NAME"; fi) >> $GITHUB_ENV
+          echo IMAGE_TAG=$(if [ -n "$GITHUB_REF_NAME" ]; then echo "$GITHUB_REF_NAME"; else echo "main-${GITHUB_SHA::7}"; fi) >> $GITHUB_ENV
```
niklasad1, line 236: This is only invoked for tags and commits so what you have done looks good to me.

.github/workflows/ci.yml
```diff
@@ -233,7 +233,7 @@ jobs:
 
       - name: Prepare Docker environment
         run: |
-          echo IMAGE_TAG=$(if [ "$GITHUB_REF" == "refs/heads/main" ]; then echo "main-${GITHUB_SHA::7}"; else echo "$GITHUB_REF_NAME"; fi) >> $GITHUB_ENV
```
niklasad1, line 236: The old thingy looks correct to me what's the issue with having commits that isn't released tagged as `main-<commit hash>`? 

I let @sergejparity answer this because I don't understand the issue most likely there is something in the deploy logic that requires a version somewhere? :)

sergejparity: @arminteimouri change you propose just flips `then-else` condition. But the result will be the same. 

Actually mine solution came from preposition that IMAGE_TAG's like `main-xxxxxxx` will happen more often and I've put it to the first place.

arminteimouri: > The old thingy looks correct to me what's the issue with having commits that isn't released tagged as `main-<commit hash>`?

check it – peep the image name, right? That's the key to knowing what version we're running on GKE. But hold up, when you glance at the code, it's like, "What version is this?" 'Cause it's all about the commit hash, and that ain't telling us squat. Sure, we can dive into the repo, track down the commit, find the tag – but why all that hassle? We can keep it chill and just make the version clear from the get-go

sergejparity: @arminteimouri probably @niklasad1 trying to tell, that this PR actually doesn't change anything apart from sequence of how `if` condition is evaluated.

it was:
```
if (push to branch `main`)
  IMAGE_TAG="main-xxxxsha"
else
 IMAGE_TAG="v0.x.x"
fi
```

it became:
```
if (tag pushed)
  IMAGE_TAG="v0.x.x"
else
 IMAGE_TAG="main-xxxxsha"
fi
```

niklasad1: Alright, I see we could change that to rename "main-xxxxsha" to "v1.3.0-unstable-<xxxsha>" instead, I reckon that would help to understand which version you want to deploy right?

The version could be parsed from https://github.com/paritytech/polkadot-staking-miner/blob/main/Cargo.toml#L3

arminteimouri: won't that tweak make the pushed Docker image name match the tag and be easy to know which release tag it's linked to? 
we got the `GITHUB_REF_NAME` in the mix, rolling as the image name.

arminteimouri: > We could rename "main-xxxxsha" to "v1.3.0-unstable-xxxxsha" instead,

That make things much clearer and easier to read.

sergejparity: > Switching up that sequence does the trick, letting the `image tag` snag the victory over the `main-commitHash` name and getting elected as the image name. Correct me please if I'm mistaken.

Not really, this job can distinguish triggering event e.g. push to the `main` branch in this case "$GITHUB_REF" == "refs/heads/main"
And second case is when `tag` is pushed.
"$GITHUB_REF" == "refs/tags/vX.X.X"

That's why I'm talking, that you just swapped checks, and they are not overlapping, so order doesn't matter.

niklasad1:
Closing this as stale and it could be useful but not really a priority for now

PR #761, opened 2024-01-04
CI improve nightly GHA workflow

sergejparity:
This PR improves `nightly` workflow by moving `Polkadot` and `Staking miner playground` binaries builds into separate workflows and just reusing its artifacts.

* `Polkadot` is rebuilt every day at 22:00

* `Staking miner playground` once a month or when changes to its codebase is made.

* Both can be triggered manually if needed


.github/workflows/build-staking-miner-playground-for-nightly.yml
```diff
@@ -0,0 +1,44 @@
...
+          cd staking-miner-playground
+          cargo build --release --features test-trimming
+          ./target/release/staking-miner-playground --version
+
+      - name: upload staking-miner-playground binary
+        uses: actions/upload-artifact@v3
+        with:
+          name: staking-miner-playground-binary
+          path: ./staking-miner-playground/target/release/staking-miner-playground
+          retention-days: 8
```
niklasad1, line 43: I think you could increase retention-rate to something like 30 days because the staking-miner playground hasn't been changed for a long time.

.github/workflows/build-staking-miner-playground-for-nightly.yml
```diff
@@ -0,0 +1,44 @@
+name: Build staking miner playground binary for nightly tests
+
+on:
+  # Run when changes to staking miner playground is made
+  push:
+    branches:
+      - main
+    paths:
+      - 'staking-miner-playground/**'
```
niklasad1, line 9: :+1:

niklasad1 (approved):
awesome thanks

Let's create a follow-up PR to move the "nightly tests" to the CI instead, it shouldn't be that slow a few minutes.

.github/workflows/build-staking-miner-playground-for-nightly.yml
sergejparity: Ok. Let it be 31, and binary will be rebuild on 1st day of every month :)

sergejparity:
> awesome thanks
> 
> Let's create a follow-up PR to move the "nightly tests" to the CI instead, it shouldn't be that slow a few minutes.

Yes. It's taking ~5 mins https://github.com/sergejparity/polkadot-staking-miner/actions/runs/7386921007

alvicsam (approved):

PR #773, opened 2024-01-16
chore: update deps in staking-miner-playground

niklasad1:
This PR updates the dependencies in staking-miner-playground and most of it is just noise.

Review `runtime/src/lib.rs` carefully and the rest is not that important.


staking-miner-playground/node/src/command_helper.rs
```diff
@@ -21,7 +21,7 @@
 
 use crate::service::FullClient;
 
-use node_template_runtime as runtime;
+use runtime;
```
niklasad1, line 24:
```suggestion
```

staking-miner-playground/runtime/src/pallet_config_block.rs
niklasad1, line 1: This is removed because the `ConfigBlock size` doesn't work when running it with subxt because it relies on the metadata and this has no effect.

The only way is to change block length and block weight is by runtime upgrade.

staking-miner-playground/runtime/src/lib.rs
```diff
@@ -155,34 +153,25 @@ parameter_types! {
 	pub const BlockHashCount: BlockNumber = 2400;
 	pub const SS58Prefix: u8 = 42;
 
-	pub ConfigurableBlockLength: limits::BlockLength = limits::BlockLength { max: PerDispatchClass::new(|_| ConfigBlock::block_length()) };
-	pub ConfigurableBlockWeight: limits::BlockWeights = limits::BlockWeights::simple_max(Weight::from_parts(ConfigBlock::block_weight(), u64::MAX));
```
niklasad1, line 159: This is removed because the `configurable block weight/length` doesn't work when running it with subxt because it relies on the metadata and this has no effect.

The only way is to change block length and block weight is by runtime upgrade.

staking-miner-playground/runtime/src/lib.rs
niklasad1, line 1: @gpestana Can you have a look at the staking configuration in this file that I didn't make any weird change?

staking-miner-playground/runtime/src/lib.rs
```diff
@@ -396,6 +382,7 @@ parameter_types! {
 	pub const RewardCurve: &'static PiecewiseLinear<'static> = &REWARD_CURVE;
 	pub const MaxNominatorRewardedPerValidator: u32 = 256;
 	pub const OffendingValidatorsThreshold: Perbill = Perbill::from_percent(17);
+	pub const MaxControllersInDeprecationBatch: u32 = 751;
```
niklasad1, line 385:
```suggestion
```

staking-miner-playground/runtime/src/lib.rs
```diff
@@ -448,17 +432,20 @@ impl pallet_staking::Config for Runtime {
...
 
 parameter_types! {
 	pub const StakingUnsignedPriority: TransactionPriority = TransactionPriority::max_value() / 2;
 
 	// signed config
 	pub const SignedRewardBase: Balance = 1 * DOLLARS;
-	pub const SignedDepositBase: Balance = 1 * DOLLARS;
+	pub const SignedFixedDeposit: Balance = 1 * DOLLARS;
 	pub const SignedDepositByte: Balance = 1 * CENTS;
+	pub const SignedDepositIncreaseFactor: Percent = Percent::from_percent(10);
```
jsdw, line 448: I don't know about this offhand; how do you know which value to set?

niklasad1: I copied it from the westend/polkadot runtime but @gpestana can you double check  :)

tadeohepperle (approved):

PR #803, opened 2024-04-03
fix: SignedPhase is removed from metadata

niklasad1:
Because the `constant SignedPhase` is removed from the metadata we have no good
alternativ than to hardcode it for now.

In addition I fetched the updated metadata from polkadot master and updated some polkadot-sdk deps.


src/commands/dry_run.rs
```diff
@@ -115,12 +116,8 @@ where
...
 		let nonce = client.rpc().system_account_next_index(signer.account_id()).await?;
 		let tx = epm::signed_solution(raw_solution)?;
-		let xt = client.chain_api().tx().create_signed_with_nonce(
-			&tx,
-			&*signer,
-			nonce,
-			Default::default(),
-		)?;
+		let params = DefaultExtrinsicParamsBuilder::new().nonce(nonce).build();
+		let xt = client.chain_api().tx().create_signed(&tx, &*signer, params).await?;
```
niklasad1, line 120: changed API in subxt

src/main.rs
```diff
@@ -209,85 +208,49 @@ async fn run_command(
 }
 
 /// Runs until the RPC connection fails or updating the metadata failed.
-async fn runtime_upgrade_task(client: Client, tx: oneshot::Sender<Error>, mut spec_version: u32) {
```
niklasad1, line 212: this bug has been fixed in subxt and we can rely on that now :)

src/client.rs
```diff
@@ -9,8 +9,6 @@ pub struct Client {
 	rpc: RpcClient,
 	/// Access to chain APIs such as storage, events etc.
 	chain_api: ChainClient,
-	/// Raw RPC client.
-	raw_rpc: RawRpcClient,
```
niklasad1, line 13: Not needed anymore

src/commands/monitor.rs
```diff
@@ -530,26 +543,19 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
...
 	at: &Header,
+	signed_phase_length: u64,
 ) -> Result<(), Error> {
 	let tx = epm::signed_solution(RawSolution { solution, score, round })?;
 
 	// TODO: https://github.com/paritytech/polkadot-staking-miner/issues/730
 	//
-	// The extrinsic mortality length is static and doesn't know when the
-	// signed phase ends.
-	let signed_phase_len = client
```
lexnv, line 540: Just to double check: Instead of fetching the signed phase constant from the chain, we define our own signed phase in the static types.rs. And this is because the constant from the chain wasn't what we'd expect?

lexnv: from the PR description: its because the `SignedPhase` is removed from westend in preparation for the merkalized metadata.

lexnv (approved):
LGTM!

src/static_types.rs
```diff
@@ -99,6 +99,9 @@ pub mod westend {
 		>(16)
 	);
 
+	// SYNC https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/westend/src/lib.rs#L451-#L453.
```
lexnv, line 102: A crazy idea would be to have a macro which keeps this in sync at compile time or a CI job that fetches that value, but it sounds a bit too complicated since the signed phase depends on multiple constants defined in substrate 🤔

jsdw (approved):
Looks sensible!

gpestana (approved):

src/static_types.rs
```diff
@@ -99,6 +99,9 @@ pub mod westend {
...
+	pub const SIGNED_PHASE_LENGTH: u64 = 150;
```
gpestana: This is a temp fix but I wonder if there are any ideas to make sure this value is up to date throughout all chains? or should we just keep a sensible upper bound since this is just used for the tx mortality?

If we leave it as is, I'd suggest us to rename `SIGNED_PHASE_LENGTH` to `SIGNED_PHASE_LENGTH_ESTIMATE` or something similar so that the const is now mistakenly used in the future.

niklasad1: it's removed now and I changed it to use `babe::epoch_duration` as upper bound/tx mortality ok?

PR #848, opened 2024-06-13
ci: staking-miner-playground check only if changed

niklasad1:


staking-miner-playground/Cargo.lock
```diff
@@ -583,7 +583,7 @@ checksum = "93f2635620bf0b9d4576eb7bb9a38a55df78bd1205d26fa994b25911a69f212f"
 dependencies = [
  "bitcoin_hashes",
  "rand 0.8.5",
- "rand_core 0.6.4",
+ "rand_core 0.5.1",
```
niklasad1, line 586:
```suggestion
 "rand_core 0.6.4",
```

PR #856, opened 2024-06-28
Fix: Feasibility check fails when all voters have the same staked amount

CertainLach:
`TrimmedVotes` uses `BTreeMap` to sort voters by stake, which results in voters with the same stake to be dropped from election.

In case if `desired_targets` == `total_voters` (on private network) it results in immediate feasibility check failure in `solution_weight`->`epm::mock_votes`, as `voters < desired_targets`.

### DIFF
--- a/src/epm.rs
+++ b/src/epm.rs
@@ -26,7 +26,8 @@ use crate::{
 };
 
 use std::{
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
nice catch, looks good to me

PR #927, opened 2024-10-31
remove `subxt substrate-compat` feature

niklasad1:
This replaces `subxt substrate compat feature` with our own `PairSigner` to avoid depending on different versions of sp-core.

### DIFF
--- a/src/signer.rs
+++ b/src/signer.rs
@@ -16,30 +16,62 @@
 
 //! Wrappers around creating a signer account.
 
-use crate::{error::Error, prelude::*};
-use sp_core::Pair as _;
+use crate::{
+	error::Error,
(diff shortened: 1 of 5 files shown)


src/signer.rs
```diff
@@ -16,30 +16,62 @@
...
+use polkadot_sdk::{
+	sp_core::Pair as PairT,
+	sp_runtime::{
+		traits::{IdentifyAccount, Verify},
+		MultiSignature as SpMultiSignature,
+	},
+};
+use subxt::Config;
+
+/// A [`Signer`] implementation that can be constructed from an [`polkadot_sdk::sp_core::Pair`].
```
niklasad1, line 32:
```suggestion
/// A [`Signer`] implementation that can be constructed from an [`Pair`].
```

PR #1008, opened 2025-04-07
multi block miner: do_reduce configurable via CLI (#995)

sigurpol:
Add the `--do-reduce` flag to the `experimental-monitor-multi-block` command.

This flag controls whether the reduction step is performed during the NPoS solution mining process for the multi-block election path.

### DIFF
--- a/src/commands/types.rs
+++ b/src/commands/types.rs
@@ -181,4 +181,7 @@ pub struct ExperimentalMultiBlockMonitorConfig {
 
     #[clap(long, value_parser, default_value = "if-leading")]
     pub submission_strategy: SubmissionStrategy,
+
+    #[clap(long, default_value_t = false)]
+    pub do_reduce: bool,
 }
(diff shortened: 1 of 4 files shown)


src/commands/types.rs
```diff
@@ -181,4 +181,7 @@ pub struct ExperimentalMultiBlockMonitorConfig {
 
     #[clap(long, value_parser, default_value = "if-leading")]
     pub submission_strategy: SubmissionStrategy,
+
+    #[clap(long, default_value_t = false)]
```
niklasad1, line 185: Cool, some docs on this? 

Like `Reduce the solution which may prevent further trimming of the solution`.

niklasad1 (approved):

src/commands/types.rs
sigurpol: > Cool, some docs on this?
> 
> Like `Reduce the solution which may prevent further trimming of the solution`.

Sure, good point. Added [here](https://github.com/paritytech/polkadot-staking-miner/pull/1008/commits/86d7da26ac7eabba412f909214bc0fd0d68fd145)

PR #1015, opened 2025-04-11
Bump subxt to `0.41.0`

sigurpol:
Driven-by:
- remove subxt-core as dependency and add `subxt-rpcs` as dependency
- fix compilation errors coming from the version bump of `subxt`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ pin-project-lite = "0.2"
 # subxt
 scale-value = "0.18"
 subxt = "0.41.0"
-subxt-core = "0.41.0"
 subxt-rpcs = "0.41.0"
 
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,16 +21,31 @@ serde = "1.0"
 serde_json = "1.0"
 futures = "0.3"
 thiserror = "2.0"
-tokio = { version = "1.44", features = ["macros", "rt-multi-thread", "sync", "signal"] }
+tokio = { version = "1.44", features = [
+    "macros",
+    "rt-multi-thread",
+    "sync",
+    "signal",
+] }
 pin-project-lite = "0.2"
 
 # subxt
 scale-value = "0.18"
-subxt = "0.40.0"
-subxt-core = "0.40.0"
+subxt = "0.41.0"
+subxt-core = "0.41.0"
+subxt-rpcs = "0.41.0"
+
 
 # polkadot-sdk
-polkadot-sdk = { git = "https://github.com/paritytech/polkadot-sdk", features = ["frame-election-provider-support", "frame-support", "sp-npos-elections", "sp-core", "sp-runtime", "pallet-election-provider-multi-phase", "pallet-election-provider-multi-block"], rev = "a92c3b2585" }
+polkadot-sdk = { git = "https://github.com/paritytech/polkadot-sdk", features = [
+    "frame-election-provider-support",
+    "frame-support",
+    "sp-npos-elections",
+    "sp-core",
+    "sp-runtime",
+    "pallet-election-provider-multi-phase",
+    "pallet-election-provider-multi-block",
+], rev = "a92c3b2585" }
 
 # prometheus
 prometheus = "0.14"
--- a/src/commands/legacy/monitor.rs
+++ b/src/commands/legacy/monitor.rs
@@ -40,10 +40,7 @@ use polkadot_sdk::{
     sp_npos_elections,
 };
 use std::sync::Arc;
-use subxt::{
-    backend::legacy::rpc_methods::DryRunResult, config::Header as _, error::RpcError,
-    Error as SubxtError,
-};
+use subxt::{backend::legacy::rpc_methods::DryRunResult, config::Header as _};
 use tokio::sync::Mutex;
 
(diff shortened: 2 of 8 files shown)


sigurpol:
Markind as draft, let me implement the From trait for converting subxt_rpcs::Error to error::Error. This will allow us to use the ? operator directly with the RPC error avoid the annoying map I have added here and there. Wait for it 😄

src/commands/legacy/monitor.rs
```diff
@@ -433,14 +430,14 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
...
-    let xt_cfg = if let Ok(len) = client
+    let xt_cfg = if let Ok(_) = client
         .chain_api()
         .constants()
         .at(&runtime::constants().babe().epoch_duration())
     {
         ExtrinsicParamsBuilder::default()
             .nonce(nonce)
-            .mortal(at, len)
+            .mortal(at.number() as u64)
```
niklasad1, line 440:
```suggestion
            .mortal(len)
```

src/commands/legacy/monitor.rs
```diff
@@ -433,14 +430,14 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
     // TODO: https://github.com/paritytech/polkadot-staking-miner/issues/730
     //
     // The extrinsic mortality length is static and it doesn't know when the signed phase ends.
-    let xt_cfg = if let Ok(len) = client
+    let xt_cfg = if let Ok(_) = client
```
niklasad1, line 433:
```suggestion
    let xt_cfg = if let Ok(len) = client
```

src/commands/legacy/monitor.rs
niklasad1: It should be mortal for n blocks,  block number is the current block number these are different.

https://docs.rs/subxt/latest/subxt/config/struct.DefaultExtrinsicParamsBuilder.html#method.mortal

src/commands/legacy/monitor.rs
```diff
@@ -455,10 +452,12 @@ async fn submit_and_watch_solution<T: MinerConfig + Send + Sync + 'static>(
     if dry_run {
         let dry_run_bytes = client.rpc().dry_run(xt.encoded(), None).await?;
 
-        match dry_run_bytes.into_dry_run_result(&client.chain_api().metadata())? {
+        match dry_run_bytes.into_dry_run_result()? {
             DryRunResult::Success => (),
             DryRunResult::DispatchError(e) => {
-                return Err(Error::TransactionRejected(e.to_string()))
+                return Err(Error::TransactionRejected(
+                    String::from_utf8_lossy(e).to_string(),
```
niklasad1, line 459: let's try to decode this to `subxt::error::DispatchError::decode_from(bytes, metadata)?;`

src/error.rs
```diff
@@ -63,3 +63,15 @@ impl From<subxt_core::Error> for Error {
         Self::Subxt(e.into())
     }
 }
+
+impl From<subxt_rpcs::Error> for Error {
+    fn from(e: subxt_rpcs::Error) -> Self {
+        Self::Other(format!("RPC error: {}", e))
```
niklasad1, line 69: Can we do `Error::Subxt(subxt::Error::Rpc(e.into()))` instead of Error::Other?

Because we are doing some special for some rpc errors.

Cargo.toml
```diff
@@ -21,16 +21,31 @@ serde = "1.0"
...
+    "signal",
+] }
 pin-project-lite = "0.2"
 
 # subxt
 scale-value = "0.18"
-subxt = "0.40.0"
-subxt-core = "0.40.0"
+subxt = "0.41.0"
+subxt-core = "0.41.0"
```
niklasad1, line 35: Can you check whether we can remove subxt-core after https://github.com/paritytech/subxt/pull/1948/files was included? 

Just try to remove the From impl here: https://github.com/paritytech/polkadot-staking-miner/blob/main/src/error.rs#L61-#L65

src/commands/legacy/monitor.rs
sigurpol: Done

src/commands/legacy/monitor.rs
sigurpol: Done (ty!)

src/commands/legacy/monitor.rs
sigurpol: done

src/error.rs
sigurpol: done

sigurpol (commented):
@niklasad1 , I've addressed your comments [here](https://github.com/paritytech/polkadot-staking-miner/pull/1015/commits/693daf7cc56eefe1827b0c4ef6febc076ecf7958). Thank you for your great review 🙇

Cargo.toml
sigurpol: done

niklasad1 (approved):
Looks good.

Hopefully we don't need to downcast the error more than this then it could make sense to add some helper function to avoid repeating that noise :)

sigurpol:
> Looks good.
> 
> Hopefully we don't need to downcast the error more than this then it could make sense to add some helper function to avoid repeating that noise :)

yes, definitely! thanks for the review!

Issue #987, opened 2025-03-04
multi block miner: different submission strategies


niklasad1:
Currently the multi-block miner tries to submit all pages at once which means that many of the extrinsics will included in the same block.

We introduce something like `-pages-per-block n` where if not set, it submits all at once, and otherwise n per block.

For now the logic is as follows:

```rust
    let mut nonce = client
        .rpc()
        .system_account_next_index(signer.account_id())
        .await?;
  
   let mut pending_blocks = Vec::new();  

   for (page, solution) in paged_raw_solution.into_iter() {
          let tx = submit(page, solution).await?;
          pending_blocks.push(tx);
          nonce += 1;
   }

   wait_for_blocks_to_be_finalized(pending_blocks).await?;
```

sigurpol:
I've tried two approaches, you can read more
- [here](https://github.com/paritytech/polkadot-staking-miner/pull/1016) to add support for chunked submissions strategies
  - testing via zombienet doesn't show any benefit with 8 / 64 pages, the chunked approach results significantly slower in my test setup
-  [here - very draft ](https://github.com/paritytech/polkadot-staking-miner/pull/1020) where instead of waiting for specific events to confirm each page submission, I poll the on-chain state with exponential backoff until we see the changes reflected in storage.
 - Again, this doesn't seem to improve things and doesn't make the code look better either.

kianenigma:
This is a good addition in case the tx-pool has some issues and we want to be mild, but true that it will not speed anything up. In fact it will only slow it down a bit. and true that indeed under normal circumstances, you should be able to submit all pages at once, and the txpool should handle them correctly.

sigurpol:
I am not opposed to merging the chunked variant as a backup plan; however, I am less enthusiastic about making it the default if we assume that we should be able to submit all pages at once... If there is consensus in having it in case of emergency, I'll finish to test carefully and submit for proper review.

sigurpol:
Ok, now [the chunked submission PR](https://github.com/paritytech/polkadot-staking-miner/pull/1016) is ready for review then.
I've tested vs zombienet with different pages and chunk size and seems to behave properly.
An example of logs for a 8-page solution with chunk size = 4 below:

```log
2025-04-15T07:01:08.328175Z  INFO polkadot-staking-miner: Score registered at block 0x949d0dc3a038e66ea3abcc33ff987e1689412a3b8a8be8f4a9e72eee5a539137
[...]

PR #1016, opened 2025-04-14
Add chunked submission strategy for multi-block solutions (#987)

sigurpol:
The current implementation submits all pages concurrently, which can be the fastest option if all submissions succeed quickly and the network or node can handle the load efficiently. However, if an early transaction fails (e.g., nonce N+1), all subsequent transactions will also fail due to the nonce gap. This leads to wasted submission attempts and potentially high load on the network or node from retransmissions.

This commit introduces support for a **chunked submission strategy**, allowing the chunk size to be specified in the experimental monitor command e.g. 

```bash
RUST_LOG="polkadot-staking-miner=trace,info" ./target/release/polkadot-staking-miner --uri ws://127.0.0.1:9966 experimental-monitor-multi-block --seed-or-path //Bob --chunk-size 4
```

If the chunk size is not specified or set to zero, we will use the fully concurrent approach. Otherwise, we will submit a chunk of N pages, wait for all of them to be successfully included in a block, and then proceed to the next chunk.

This approach aims to balance speed and robustness, but it requires careful validation in real use cases. Note that if the chunk size is set to 1, we will have a fully sequential solution: submit one page, wait, then submit the next page, and so on, which may significantly slow down the process.

Testing against a simple `zombienet` setup with an 8-page solution:

| Submission Strategy | Time (seconds) |
|---------------------|----------------|
| Fully concurrent     | ~77s           |
| Chunked (size=4)    | ~105s          |


sigurpol:
@niklasad1 , @kianenigma  PTAL  when/if you have time 🙏 🙇

src/dynamic/multi_block.rs
```diff
@@ -479,12 +501,127 @@ pub(crate) async fn inner_submit_pages<T: MinerConfig + Send + Sync + 'static>(
...
+    T::Pages: Send + Sync + 'static,
+{
+    assert!(chunk_size > 0, "Chunk size must be greater than 0");
+
+    let mut failed_pages = Vec::new();
+    let mut submitted_pages = HashSet::new();
+    let total_pages = paged_raw_solution.len();
+
+    // Process pages in chunks
+    for chunk_start in (0..total_pages).step_by(chunk_size) {
```
niklasad1, line 582: you could perhaps use https://doc.rust-lang.org/std/primitive.slice.html#method.chunks here instead of doing it yourself.

src/dynamic/multi_block.rs
```diff
@@ -414,23 +430,28 @@ pub(crate) async fn submit<T: MinerConfig + Send + Sync + 'static>(
...
-pub(crate) async fn inner_submit_pages<T: MinerConfig + Send + Sync + 'static>(
+/// Helper function to submit a batch of pages and wait for their inclusion in blocks
+async fn submit_pages_batch<T: MinerConfig + Send + Sync + 'static>(
     client: &Client,
     signer: &Signer,
-    paged_raw_solution: Vec<(u32, T::Solution)>,
+    pages_to_submit: Vec<(u32, T::Solution)>,
     listen: Listen,
-) -> Result<Vec<u32>, Error> {
+) -> Result<(Vec<u32>, HashSet<u32>), Error>
```
niklasad1, line 439: Both submitted pages and failed pages are returned mostly for logging purposes?

It would be use with a separate type for this like `SubmissionResult/Status` or something like that wraps that

src/commands/multi_block/monitor.rs
```diff
@@ -165,14 +166,15 @@ async fn process_block<T>(
     submit_lock: Arc<Mutex<()>>,
     submission_strategy: SubmissionStrategy,
     do_reduce: bool,
+    chunk_size: usize,
 ) -> Result<(), Error>
 where
     T: MinerConfig<AccountId = AccountId> + Send + Sync + 'static,
-    T::Solution: Send,
-    T::Pages: Send,
+    T::Solution: Send + Sync + 'static,
```
niklasad1, line 173: do you know why these extra trait bounds are needed?

src/dynamic/multi_block.rs
sigurpol: TIL. Done [here](https://github.com/paritytech/polkadot-staking-miner/pull/1016/commits/a5d4bd98477d7d67c1098e0d94fa3b6b9ffa64be)

src/dynamic/multi_block.rs
sigurpol: Good point. Done [here](https://github.com/paritytech/polkadot-staking-miner/pull/1016/commits/ab6c8cc867021efb1413cad3a72edb167eff424f)

niklasad1 (approved):
looks good :+1:

src/commands/multi_block/monitor.rs
sigurpol: I think it was a leftover from some intermediate version. Should be better now with this [commit](https://github.com/paritytech/polkadot-staking-miner/pull/1016/commits/6ca18ea16eeb4fc34115ada5b5284e4022c9ff70)

Issue #1014, opened 2025-04-11
[documentation] No support for multiple miners running with the same account


sigurpol:
Specify in the documentation that we don't support running multiple instances of the miner with the same account. Today it would mess up with nonce since to submit the solution we get the on-chain nonce for a given user so there is a risk of collisions if multiple miners run with the same account.

sigurpol:
Documentation updated in https://github.com/paritytech/polkadot-staking-miner/pull/1021

PR #1021, opened 2025-04-15
doc: no multiple instances of the miner with the same account

sigurpol:
Update the documentation to explicitly state that running multiple instances of the miner with the same account is not allowed. This is to prevent interference with the submission mechanism, which relies on the on-chain noise for each user.

Related issue [here](https://github.com/paritytech/polkadot-staking-miner/issues/1014)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,14 @@ it's not possible to know in advance which runtimes it will work with.
 Thus, it's important to subscribe to releases to this repo or
 add some logic that triggers an alert once the polkadot-staking-miner crashes.
 
+## Important Warning
+
+**Do not run multiple instances of the miner with the same account.**
+
+The miner uses the on-chain nonce for a given user to submit solutions, which can lead to nonce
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,10 @@ it's not possible to know in advance which runtimes it will work with.
 Thus, it's important to subscribe to releases to this repo or
 add some logic that triggers an alert once the polkadot-staking-miner crashes.
 
+## Important Warning
+
+**Do not run multiple instances of the miner with the same account.** The miner uses the on-chain nonce for a given user to submit solutions, which can lead to nonce collisions if multiple miners are running with the same account. This can cause transaction failures and potentially result in lost rewards or other issues.
```
niklasad1, line 37: you could perhaps split this on several lines make it easier to read.

niklasad1 (approved):

README.md
sigurpol: done

PR #1029, opened 2025-04-22
feat: reclaim deposit for discarded solutions

### LINKED ISSUE #980: multi block miner: reclaim deposit
If a miner submits something, and it is later on it is not the best solution that we verify, we call this being discarded. In such cases, the deposit is not returned, until you call a transaction that manually cleans up the state 

https://github.com/paritytech/polkadot-sdk/pull/7601/commits/3063a057f5924ceb38a0ba6b8631c5a507ddde54#diff-59253e064f078ede8607eef3b9b0c0ed1b6301318d397eabfac5965b05ad0c50

something to sort out after we have the happy path working on westend/ah-next, needed for kusama

sigurpol:
This commit introduces logic to track submitted rounds and automatically call the clear_submission function (which dispatches the `clear_old_round_data` extrinsic) for any solution that is determined to be discarded (i.e., not the winning solution) in **past rounds**. 
The monitor now checks **all previously submitted rounds** on each block, and if a better solution has been validated or the round has ended and our submission still exists, it triggers the clearing process to reclaim the deposit.

This ensures that locked deposits for discarded solutions are properly returned to the miner without manual intervention.

Fix #980.


sigurpol:
@niklasad1 , @kianenigma  PTAL 🙏

src/commands/multi_block/monitor.rs
```diff
@@ -201,10 +226,41 @@ where
         _ => return Ok(()),
     }
 
-    // 2. If the solution has already been submitted, nothing to do
-    if has_submitted(&storage, round, signer.account_id(), n_pages).await? {
+    // 2. If the solution has already been submitted:
+    // 2.1 Check local tracking first
+    if submitted_rounds.lock().await.contains_key(&round) {
```
niklasad1, line 231: Because we support both best blocks and finalized blocks at moment it's possible that a block can be reverted and this locally `submitted_round` may not be accurate with the state on chain then.

Finalized blocks can't be reverted except hard forks IIRC, so to support such as functionality we would need to remove support for listening to best blocks which may simplify our life a bit.

That's the reason why we are checking the state at latest head like:

```rust
if has_submitted(
        &utils::storage_at_head(&client, listen).await?
)
```

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +638,210 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+    }
+
+    log::debug!(
+        target: LOG_TARGET,
+        "Checking status of PAST rounds: {:?}",
+        rounds_to_potentially_clear.iter().map(|(r,_,_)| r).collect::<Vec<_>>()
+    );
+
+    let get_finalized_storage = async {
+        let finalized_hash = client.rpc().chain_get_finalized_head().await?;
```
niklasad1, line 686: you can use utils::storage_at_head instead of this

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +638,210 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+        Ok(s) => s,
+        Err(e) => {
+            log::warn!(target: LOG_TARGET, "Failed to get finalized storage state for checking old rounds: {:?}. Skipping check.", e);
+            return Ok(());
+        }
+    };
+
+    let mut rounds_to_remove = Vec::new();
+
+    for (round_to_check, n_pages, submitted_score) in rounds_to_potentially_clear {
```
niklasad1, line 699: this should most likely only be one round to clear otherwise it's a bug?

src/commands/multi_block/monitor.rs
sigurpol: Good point. What are you suggesting as best option?
1. Only list to finalized blocks
vs
2. always double-check with chain state i.e. use local tracking as now but then validate against chain state e.g.

```rust
// Check local tracking first, but validate against chain state
if submitted_rounds.lock().await.contains_key(&round) {
    // Verify with chain state before trusting local tracking
    let storage_head = utils::storage_at_head(&client, listen).await?;
    if has_submitted(&storage_head, round, signer.account_id(), n_pages).await? {
        // Confirmed both locally and on-chain
        return Ok(());
    } else {
        // Local tracking is wrong, remove it and continue
        untrack_round_submission(
            &submitted_rounds,
            round,
            Some("Removing inconsistent local tracking for round {}")
        ).await;
        // Continue with submission attempt
    }
}
```

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +638,210 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+            )
+            .await;
+        }
+    }
 
     Ok(())
 }
+
+async fn collect_past_rounds_to_clear(
+    submitted_rounds: &Arc<Mutex<HashMap<u32, (u32, ElectionScore)>>>,
```
niklasad1, line 787: deref impl of Arc is just &Mutex no need to for &Arc<Mutex<....>

```suggestion
    submitted_rounds: &Mutex<HashMap<u32, (u32, ElectionScore)>>>,
```

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +638,210 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+            message,
+            round,
+            n_pages,
+            score
+        );
+    }
+}
+
+/// Helper method to safely remove a round's submission data from the tracking map
+async fn untrack_round_submission(
```
niklasad1, line 829: I would personally want a separate type/abstraction for the RoundSubmission with these functions as methods such that we can hide a bit of mutex details etc to make the code a bit more readable.

```rust
struct RoundSubmission(Arc<Mutex<HashMap<u32, (u32, ElectionScore)>>>);


impl RoundSubmission {
    // Insert a new submission (it may replace a previous one if block was reverted)
    fn insert(&self, round: u32, n_pages: u32, score: ElectionScore) -> { ... }
    // Collect all previous submissions 
    fn collect(&self) -> HashMap<u32, (u32, ElectionScore)> { ... }
}
```

src/commands/multi_block/monitor.rs
```diff
@@ -287,38 +395,74 @@ where
...
-                return Ok(());
+            } else {
+                // Score mismatch, revert previous submission and submit a new one.
+                log::warn!(target: LOG_TARGET, "Mined solution score {:?} differs from incomplete submission score {:?} for round {}. Bailing previous submission.", submitted_score, s.score(), round);
+                bail(listen, &client, signer.clone()).await?;
+                // Clear local tracking as we are bailing
+                untrack_round_submission(
+                    &submitted_rounds,
+                    round,
+                    Some("Removed previous submission for round {} after bailing"),
```
niklasad1, line 436: an enum for this would be neat instead of Option<&str>

src/commands/multi_block/monitor.rs
```diff
@@ -104,27 +111,43 @@ where
             .ok_or(Error::AccountDoesNotExists)?;
         prometheus::set_balance(account_info.data.free as f64);
 
+        let current_block_round = state.round;
+
+        // This checking logic runs on EVERY block
+        let rounds_to_potentially_clear =
+            collect_past_rounds_to_clear(&submitted_rounds, current_block_round).await;
+
+        check_and_clear_discarded_submissions(
```
niklasad1, line 120: you could re-use the storage from [the state](https://github.com/paritytech/polkadot-staking-miner/blob/main/src/commands/multi_block/types.rs#L139)  instead of re-fetching inside this function

src/commands/multi_block/monitor.rs
niklasad1: The entire point with best blocks is that it's faster and you may win against finalized blocks but it may bite you in the ass and if something got included in block but never got finalized or reverted.

Personally I don't think it's worth the effort of maintaining such annoyances when we are just using finalized when we are running the miner. My cents is to remove best block listening completely.

This thingy would be annoying to implement or simply for best blocks these weird stuff could happen users that uses best blocks simply has to accept the fact that it is unreliable.

/cc @kianenigma what's your take on best blocks?

src/commands/multi_block/monitor.rs
sigurpol: My understanding was that `utils::storage_at_head` gets storage at the latest best block but here we want to check past submissions here so we need to use the finalised storage for that.

src/commands/multi_block/monitor.rs
sigurpol: but it's possible to have multiple rounds in theory e.g. after the miner restarts... or so I thought :)

src/commands/multi_block/monitor.rs
sigurpol: Done

src/commands/multi_block/monitor.rs
sigurpol: I wouldn't be against to just go with finalized blocks and check the impact on the overall performance of the miner. I'll make a commit on top, if you agree

sigurpol:
@niklasad1 , @kianenigma , CI/run test failures will be addressed once https://github.com/paritytech/polkadot-sdk/pull/8310 is merged into polkadot-sdk, so please don't pay too much attention to it.

src/commands/multi_block/monitor.rs
sigurpol: Changes to support **ONLY** finalized blocks in [this commit](https://github.com/paritytech/polkadot-staking-miner/pull/1029/commits/98b9b0960b54669d2705b9301a7c1dfff74ec42c)

src/commands/multi_block/monitor.rs
```diff
@@ -22,9 +22,117 @@ use polkadot_sdk::{
...
+enum RoundUntrackingEvent {
+    Cleared, // Removed after clearing
+    Bailed,  // Removed after bailing
+    Failed,  // Removed after a failed submission
+}
+
+/// A dedicated type to manage submitted rounds tracking with clean APIs
+/// that hide the mutex implementation details
+#[derive(Clone)]
+struct RoundSubmission(Arc<Mutex<HashMap<u32, (u32, ElectionScore)>>>);
```
niklasad1, line 52: nice I like the new the new type :)

I would personally move it `multi-block/types.rs` to make this file a bit shorter and easier to read but up to you :)

src/commands/multi_block/monitor.rs
```diff
@@ -237,93 +421,145 @@ where
...
                     )
-                    .await?;
+                    .await
                 } else {
                     dynamic::inner_submit_pages_chunked::<T>(
                         &client,
                         &signer,
                         missing_pages,
-                        listen,
+                        Listen::Finalized,
```
niklasad1, line 491: remove `Listen` as well from `dynamic::inner_submit_pages_chunked` it's only used by experimental multi block stuff

src/commands/multi_block/monitor.rs
```diff
@@ -22,9 +22,117 @@ use polkadot_sdk::{
...
+        }
+
+        removed
+    }
+
+    /// Collect past rounds that should be cleared
+    async fn collect_past_rounds(&self, current_block_round: u32) -> Vec<(u32, u32)> {
+        let rounds = self.0.lock().await;
+
+        rounds
```
niklasad1, line 123: why not remove "old data" right away here?

src/commands/multi_block/monitor.rs
```diff
@@ -201,10 +330,41 @@ where
         _ => return Ok(()),
     }
 
-    // 2. If the solution has already been submitted, nothing to do
-    if has_submitted(&storage, round, signer.account_id(), n_pages).await? {
+    // 2. If the solution has already been submitted:
```
niklasad1, line 333: I would move this 2.1 and 2.2 to a separate function to make it a bit more readable and it's duplicated below, WDYT?

src/commands/multi_block/monitor.rs
```diff
@@ -22,9 +22,117 @@ use polkadot_sdk::{
...
+        self.0.lock().await.contains_key(&round)
+    }
+
+    /// Insert a new round submission into the tracking map
+    async fn insert(
+        &self,
+        round: u32,
+        n_pages: u32,
+        score: ElectionScore,
+        event: Option<RoundTrackingEvent>,
```
niklasad1, line 71: I would add an additional enum variant to be explicit here but I don't think we need it.

`None` should be replaced RoundTracking::Submitted I reckon, it could probably happen if the miner gets restarted during the sign phase.

src/commands/multi_block/monitor.rs
```diff
@@ -22,9 +22,117 @@ use polkadot_sdk::{
...
+                message,
+                round,
+                n_pages,
+                score
+            );
+        }
+    }
+
+    /// Remove a round submission from the tracking map
+    async fn remove(&self, round: u32, event: Option<RoundUntrackingEvent>) -> bool {
```
niklasad1, line 97: remove option here as well.

src/commands/multi_block/monitor.rs
```diff
@@ -22,9 +22,117 @@ use polkadot_sdk::{
...
+        round: u32,
+        n_pages: u32,
+        score: ElectionScore,
+        event: Option<RoundTrackingEvent>,
+    ) {
+        let mut rounds = self.0.lock().await;
+        rounds.insert(round, (n_pages, score));
+
+        if let Some(event) = event {
+            let message = match event {
```
niklasad1, line 77: not that important, but you could make this a display impl on the type itself to make method impl slightly cleaner.

```rust
impl std::fmt::Display for RoundTrackingEvent {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            RoundTrackingEvent::Found => write!(f, "Found submission for"),
            RoundTrackingEvent::Submitted => write!(f, "Successfully submitted and recorded""),
            RoundTrackingEvent::SubmittedPartial => write!(f, "Completed partial submission for"),
            RoundTrackingEvent::Cleared => write!(f, "Cleared discarded submission for"),
            RoundTrackingEvent::FailedSubmission => write!(f, "Failed submission for"),
        }
    }
}
```

src/commands/multi_block/monitor.rs
```diff
@@ -409,21 +673,57 @@ async fn get_submission(
...
-    storage: &Storage,
-    round: u32,
-    who: &subxt::config::substrate::AccountId32,
-    n_pages: u32,
-) -> Result<bool, Error> {
-    match get_submission(storage, round, who, n_pages).await? {
-        CurrentSubmission::Done(_) => Ok(true),
-        _ => Ok(false),
+/// Bail out of the current round submission.
+async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Error> {
```
niklasad1, line 677:
```suggestion
async fn bail(client: &Client, signer: Signer) -> Result<(), Error> {
```

src/commands/multi_block/monitor.rs
```diff
@@ -237,93 +421,145 @@ where
...
+            if s.score() == submitted_score {
+                log::info!(target: LOG_TARGET, "Found incomplete submission for round {} with matching score {:?}. Submitting missing pages.", round, submitted_score);
                 let mut missing_pages = Vec::new();
-
                 for page in s.get_missing_pages() {
-                    let solution = paged_raw_solution.solution_pages[page as usize].clone();
-                    missing_pages.push((page, solution));
+                    // Clone only needed pages
+                    if let Some(solution_page) =
+                        paged_raw_solution.solution_pages.get(page as usize)
```
niklasad1, line 468: this should be infallible, page idx must be valid I would be okay with an unwrap/expect here but fair enough

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +604,162 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+        }
+    }
 
     Ok(())
 }
+
+/// Clears the miner's submission data for a specific round to reclaim the deposit.
+/// This should be called when a previously submitted solution is known to be discarded.
+async fn clear_submission(
+    listen: Listen,
```
niklasad1, line 626: remove

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +604,162 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+            .fetch(
+                &runtime::storage()
+                    .multi_block_signed()
+                    .submission_metadata_storage(round_to_check, signer.account_id()),
+            )
+            .await?;
+
+        if maybe_submission_metadata.is_none() {
+            log::debug!(target: LOG_TARGET, "Submission metadata for past round {} gone. Removing from tracking.", round_to_check);
+            rounds_to_remove.push(round_to_check);
```
niklasad1, line 712: remove directly instead of pushing to a vec?

src/commands/multi_block/monitor.rs
```diff
@@ -237,93 +283,139 @@ where
...
+    let storage_after_mining = storage_at_finalized(&client).await?;
+    match get_submission(&storage_after_mining, round, signer.account_id(), n_pages).await? {
         CurrentSubmission::Done(score) => {
             // We have already submitted the solution with a better score or equal score
-            if !score_passes_strategy(paged_raw_solution.score, score, submission_strategy) {
+            if !score_passes_strategy(submitted_score, score, submission_strategy) {
+                log::info!(target: LOG_TARGET, "Mined solution score {:?} not better than already submitted score {:?} for round {}. Skipping.", submitted_score, score, round);
+                // Ensure it's recorded locally if somehow missed before
+                if !submitted_rounds.contains(round).await {
+                    submitted_rounds.insert(round, n_pages, score, None).await;
```
niklasad1, line 310: I think it would make more sense to use RoundTracking::Submitted here

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +604,162 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+            rounds_to_remove.push(round_to_check);
+            continue;
+        }
+
+        // For past rounds, we assume they've already progressed beyond active phases
+        // If submission metadata still exists, it needs clearing
+        let should_clear = true;
+        let clear_reason = "Past round with still existing submission metadata";
+
+        if should_clear {
```
niklasad1, line 721: needless if, it always true

src/commands/multi_block/monitor.rs
sigurpol: Done

src/commands/multi_block/monitor.rs
```diff
@@ -437,9 +604,162 @@ async fn bail(listen: Listen, client: &Client, signer: Signer) -> Result<(), Err
...
+            log::info!(target: LOG_TARGET, "Submission for past round {} detected as DISCARDED (Reason: {}). Attempting to clear.", round_to_check, clear_reason);
+            match clear_submission(
```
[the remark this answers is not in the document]
sigurpol: Proposed fix [here](https://github.com/paritytech/polkadot-staking-miner/pull/1029/commits/1a941b110417a69625b903ac0a2d53f15bb37964)

src/commands/multi_block/monitor.rs
sigurpol: I think you are right and now [here](https://github.com/paritytech/polkadot-staking-miner/pull/1029/commits/ab70b5d07b4526b10ca6575966b1e14798fd0fa1) we handle clearing submission only for the most recent past round.

src/commands/multi_block/monitor.rs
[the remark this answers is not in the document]
sigurpol: Now I am wondering if, instead of the current approach, we could keep all past rounds and related submissions while having a queue of requests to clear. We could also have a dedicated thread to process them until the queue is empty. If it's empty, we can pause. @niklasad1, what do you think?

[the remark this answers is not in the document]
sigurpol: Here is the queue-based approach -> https://github.com/paritytech/polkadot-staking-miner/pull/1029/commits/c413cc035f09f34227c3b7c6df6e22914ad3cf56

src/commands/multi_block/monitor.rs
```diff
@@ -148,11 +228,119 @@ where
...
+    notify: Arc<Notify>,
+) -> Result<(), Error> {
+    // Initial queue size update
+    prometheus::set_clearing_round_queue_size(clear_receiver.len());
+
+    loop {
+        // Process any rounds in the queue
+        let mut processed = false;
+
+        while let Ok(round_to_clear) = clear_receiver.try_recv() {
```
niklasad1, line 246: this will essentially poll the receiver, let's the async API to avoid waking up the task that often when there is nothing to consume

```suggestion
        while let Ok(round_to_clear) = clear_receiver.recv().await {
```

sigurpol:
Thanks for the comment, @kianenigma. I see your valid points. Let me explore your approach and see how it would look. I'm also interested in @niklasad1's opinion on this!

src/commands/multi_block/monitor.rs
sigurpol: Done a variant of this in https://github.com/paritytech/polkadot-staking-miner/pull/1029/commits/4a7899a23782f711372152ac6c756c80bcb3e70c

niklasad1:
> and the fact that you keep your internal state of what you have submitted, is my main point of disagreement. I would have done this simpler:

> 1. Keep the event loop of monitor-and-submit as separate as possible. This thread should always look at the current round, if Phase::Signed mine something and submit it. Its responsibilities end here. It won't keep track of what has submitted or not
 > 2. A totally separate thread is always looking at the SignedSubmissions storage, and checking if there are any data starting with key (x, our_account) where x < current_round in the storage, and clean that up. This thread can run every block, every 10 blocks, or every day, it doesn't need to be super on-time.

It seems like a good idea to me because this solution became a bit complicated and I don't reconnecting/restarting shouldn't take that long (tens of seconds is my guess) so the probability missing the round transition should quite low but possible.

There is also a possibility to use the reconnecting-rpc-client from subxt to protect against this stuff as well, but still possible that it crashes for some other reason.

I would think this solution would be a good first step and then improve it if really losing submissions is a big issue then we could write them to a file or something to re-use on start-up.

sigurpol:
> > and the fact that you keep your internal state of what you have submitted, is my main point of disagreement. I would have done this simpler:
> 
> > 1. Keep the event loop of monitor-and-submit as separate as possible. This thread should always look at the current round, if Phase::Signed mine something and submit it. Its responsibilities end here. It won't keep track of what has submitted or not
> > 2. A totally separate thread is always looking at the SignedSubmissions storage, and checking if there are any data starting with key (x, our_account) where x < current_round in the storage, and clean that up. This thread can run every block, every 10 blocks, or every day, it doesn't need to be super on-time.
> 
> It seems like a good idea to me because "the current implementation in this PR" became a bit complicated and I don't think reconnecting/restarting shouldn't take that long (tens of seconds is my guess) so the probability missing the round transition should quite low but possible.
> 
> There is also a possibility to use the [reconnecting-rpc-client from subxt](https://github.com/paritytech/polkadot-staking-miner/issues/1033) to protect against this stuff as well, but still possible that it stops for some other reason.
> 
> I would think this Kian's suggested would be a good first step and then improve it if really losing submissions is a big issue then we could write them to a file or something to re-use on start-up.
> 
> but really it's to you @sigurpol :)

Thanks @kianenigma and @niklasad1 . I do agree with @kianenigma 's suggestion actually. I think the pros are pretty strong:

- Chain as source of truth: no longer duplicate state that's already tracked on-chain
- Crash resistance: If the miner crashes and restarts, it will still find and clear old submissions
- Cleaner separation of concerns: The main monitor process focuses solely on current round submissions
- Simplicity: No need for complex coordination between processes with channels and notifications

so seems a clear winner 🚀

sigurpol:
Feature has been implemented in #1095

PR #1035, opened 2025-04-25
Track SDK's master instead of ankn/staking-async-3

### LINKED ISSUE #1013: [Cleanup] Remove reference to polkadot-sdk branch `ankn/staking-async-3` in Cargo.toml
With https://github.com/paritytech/polkadot-staking-miner/pull/1012 we have updated the miner to properly support multi-block stacking vs specific polkadot-sdk rev `a92c3b2585`.
Then with #1023 we have pointed directly to `ankn/staking-async-3` branch.

Once `ankn/staking-async-3` gets merged , update  `Cargo.toml` to point to sdk master branch and not to a specific branch / rev. Or even better, as per Niklas's comment, once it's released on crates.io we can do proper release on crates.io, https://crates.io/crates/polkadot-staking-miner

**NOTE**: A follow-up task will be then to remove all legacy code once we go all in with the multi-block approach. And rename `experimental-xx` in just `monitor`.

sigurpol:
Fix #1013.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -869,7 +869,7 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 [[package]]
 name = "asset-test-utils"
 version = "7.0.0"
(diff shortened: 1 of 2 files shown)


sigurpol:
Awesome even better :)

PR #1040, opened 2025-04-28
tests: add e2e tests for multi-block miner

niklasad1:


tests/multi_block_monitor.rs
```diff
@@ -0,0 +1,185 @@
...
+
+            // Score registration.
+            if let Some(item) = ev.as_event::<Registered>()? {
+                if item.1 == alice() {
+                    score_submitted = true;
+                }
+            }
+
+            // Page submission.
+            if let Some(item) = ev.as_event::<Stored>()? {
```
niklasad1, line 121: TODO: we could also look for the rewarded event

tests/multi_block_monitor.rs
```diff
@@ -0,0 +1,185 @@
...
+        node_cmd.stdout.take().unwrap(),
+        node_cmd.stderr.take().unwrap(),
+        tx,
+    );
+
+    while let Some(line) = rx.recv().await {
+        log::info!("{}", line);
+
+        // We regard this as zombienet started and we can start the miner.
+        if line.to_lowercase().contains("node information") {
```
niklasad1, line 171: TODO: we could also parse the rpc port from zombienet CLI output but I think it's fine to assume that it start up at port 9966 but if this port is occupied polkadot-sdk will assign a random one.

tests/multi_block_monitor.rs
sigurpol: Added check on rewarded too

sigurpol (approved):

sigurpol:
Let's merge in order to ensure all needed artifacts (binaries, chainspec) are in place and let's improve / fix / iterate.

PR #1052, opened 2025-05-15
Switch jobs to parity-large

sigurpol:
This should fix the issue we were having with the public available runner with just 14GB of disk space, resulting in jobs failing due to no space left.


sigurpol:
@seadanda , @kianenigma  if you have time....

seadanda (approved):

PR #1056, opened 2025-05-21
CI/CD: more verbose logs for EPMB and staking

sigurpol:
We have seen the multi-block integration test failing once so far in CI/CD (see #1055). If it happens again, let's make sure to have enough info on EPMB/staking side.
Let's also make the miner itself more verbose.

NOTE: in order to be able to build the docker image via CI, we also revert #1052  and use again `ubuntu-latest` as runner, hoping not to hit again the `no disk space left` issue.

### DIFF
--- a/zombienet-staking-runtimes.toml
+++ b/zombienet-staking-runtimes.toml
@@ -23,5 +23,5 @@ chain_spec_path = "/github/home/.config/polkadot-staking-miner/parachain.json"
 name = "charlie"
 rpc_port = 9966
 args = [
-	"-lruntime::system=debug,runtime::multiblock-election=debug,runtime::staking=debug,runtime::staking::rc-client=trace",
+	"-lruntime::system=debug,runtime::multiblock-election=trace,runtime::staking=info,runtime::staking::rc-client=trace",
 ]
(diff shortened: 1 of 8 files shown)


zombienet-staking-runtimes.toml
```diff
@@ -23,5 +23,5 @@ chain_spec_path = "/github/home/.config/polkadot-staking-miner/parachain.json"
 name = "charlie"
 rpc_port = 9966
 args = [
-	"-lruntime::system=debug,runtime::multiblock-election=debug,runtime::staking=debug,runtime::staking::rc-client=trace",
+	"-lruntime::system=debug,runtime::multiblock-election=trace,runtime::staking=info,runtime::staking::rc-client=trace",
```
Ank4n, line 26: did you mean to decrease verbosity for staking?

sigurpol: > did you mean to decrease verbosity for staking?

good catch, thank you. No, I didn't mean to. Fixed [here](https://github.com/paritytech/polkadot-staking-miner/pull/1056/commits/1e41330444687ecc51b93c7bde2ed7595a1d3f39)

sigurpol: NOTE: unrelated to this change, but for the issue on building the docker image (e.g. [here](https://github.com/paritytech/polkadot-staking-miner/actions/runs/15160791221/job/42630795497)) after we have switched to `parity-large` in #1052 , I've pinged @alvicsam / CI 🧙‍♂️  on Element asking for help on the better way to handle the issue. It's not specific to this PR so can be part of another one but we have to fix the issue to fully unlock CI for the miner. Reverting #1052 is not an option either since CI was failing (not enough disk space)

seadanda (approved):

PR #1057, opened 2025-05-21
ci: test docker

alvicsam:
PR out of curiosity why the docker doesn't build on self-hosted runners


sigurpol:
@alvicsam  do you want to keep this PR alive or can we close it? We have switched back to ubuntu-latest for most of the jobs and it's weeks we have no problem with CI

PR #1064, opened 2025-05-30
static config: Add support for AH and staking-async chains

sigurpol:
- Deprecated `asset-hub-next`, replaced by `staking-async-parachain` and `staking-async-rc` in
https://github.com/paritytech/polkadot-sdk/pull/8422
- Add support for `statemint`, `statemine`, `westmint` AH chains, without dropping the polkadot / kusama / westend RC variants until we deprecate the legacy `monitor` path and rely only on EPMB

For the time being, static miner's config for `StakingAsync` is a copy-paste of `Westend` one but will evolve independently.

As a separate topic (see #994), a partial / complete removal of the static configuration per chain will be tackled.

### DIFF
--- a/src/opt.rs
+++ b/src/opt.rs
@@ -51,7 +51,7 @@ pub enum Chain {
     Kusama,
     Polkadot,
     SubstrateNode,
-    AssetHubNext,
+    StakingAsync,
 }
 
 impl fmt::Display for Chain {
@@ -61,7 +61,7 @@ impl fmt::Display for Chain {
             Self::Kusama => "kusama",
             Self::Westend => "westend",
             Self::SubstrateNode => "node",
-            Self::AssetHubNext => "asset-hub-next",
+            Self::StakingAsync => "staking-async",
         };
         write!(f, "{}", chain)
     }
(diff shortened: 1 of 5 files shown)


src/static_types/legacy.rs
```diff
@@ -314,3 +314,68 @@ pub mod kusama {
...
+            VoterIndex = u32,
+            TargetIndex = u16,
+            Accuracy = sp_runtime::PerU16,
+            MaxVoters = ConstU32::<22500>
+        >(16)
+    );
+
+    #[derive(Debug)]
+    pub struct MinerConfig;
+    impl pallet_election_provider_multi_phase::unsigned::MinerConfig for MinerConfig {
```
kianenigma, line 335: for staking async, we only need to implement `multi_block::MinerConfig`.

src/static_types/legacy.rs
```diff
@@ -314,3 +314,68 @@ pub mod kusama {
         }
     }
 }
+
+/// This is used to test against staking-async runtimes from the SDK.
+pub mod staking_async {
```
kianenigma, line 319: I don't think this is needed -- in staking-async, we only have and support the multi-block stuff.

src/opt.rs
```diff
@@ -73,9 +73,13 @@ impl std::str::FromStr for Chain {
...
         match s {
             "polkadot" => Ok(Self::Polkadot),
+            "statemint" => Ok(Self::Polkadot), // Polkadot AH
             "kusama" => Ok(Self::Kusama),
+            "statemine" => Ok(Self::Kusama), // Kusama AH
             "westend" => Ok(Self::Westend),
-            "asset-hub-next" => Ok(Self::AssetHubNext),
+            "westmint" => Ok(Self::Westend), // Westend AH
+            "staking-async-parachain" => Ok(Self::StakingAsync),
+            "staking-async-rc" => Ok(Self::StakingAsync),
```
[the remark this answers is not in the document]
sigurpol: true, removed

src/static_types/legacy.rs
sigurpol: good point, removed

PR #1076, opened 2025-06-06
Use subxt's reconnecting RPC client, perform an early exit check on the phase, and skip mining if the submission lock is already taken.

### LINKED ISSUE #1075: Miner on the Westend Asset Hub hangs during a long-running session.
While testing the miner against `WAH` via

```bash
RUST_LOG="polkadot-staking-miner=trace,subxt=debug,info" ./target/release/polkadot-staking-miner --uri wss://westend-asset-hub-rpc.polkadot.io experimental-monitor-multi-block --seed-or-path $MY_SEED
```

I was able to reproduce few times a case where the miner is stalled. 
From the logs
```bash
2025-06-05T16:42:04.840686Z TRACE polkadot-staking-miner: Processing block=11876012 round=15, phase=Off
2025-06-05T16:42:08.963686Z TRACE polkadot-staking-miner: Processing block=11876013 round=15, phase=Off
2025-06-05T16:42:33.175583Z TRACE polkadot-staking-miner: Processing block=11876014 round=15, phase=Off
2025-06-05T16:42:41.037520Z TRACE polkadot-staking-miner: Processing block=11876015 round=15, phase=Off
```
and then nothing is printed out for minutes (forever?).

So far, I have been able to reproduce the issue only while testing on `WAH`, not in my local zombienet setup. 
The issue appears to occur during the `Off` phase, but this is merely speculation, so please take it with a grain of salt.

sigurpol:
- Add automatic RPC reconnection support with exponential backoff using subxt's `reconnecting-rpc-client` feature. This makes the miner more resilient to temporary network issues and RPC disconnections.

- Added an early exit check on the phase to avoid unnecessary RPC calls.

Monitoring is optimized by checking the phase before making additional storage queries when they are not needed. If the phase is different from Signed or Snapshot, the miner has no tasks to perform, making further access to storage and multiple expensive RPC calls unnecessary.

- Automatically revert of incomplete submissions if the phase changes from  Signed during the process  

- Skipped mining if the submission lock is already taken.

This enhances mining concurrency by skipping block processing if another task has already acquired the lock, rather than blocking until it becomes available.
Now, when multiple consecutive blocks are in the `Signed` phase for the same round:

1. **Block N**: Acquires the submission lock and proceeds with mining/submission.
2. **Block N+1, N+2, etc.**: Attempts `try_lock()`, fails immediately, logs a trace message, and returns as NO-OP.

This eliminates redundant tasks such as:
   - Waiting for the submission lock
   - Fetching storage again
   - Checking submission status
   - Only to realize the work is already done.
   
   Close #1075

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3030,6 +3030,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "finito"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2384245d85162258a14b43567a9ee3598f5ae746a1581fb5d3d2cb780f0dbf95"
+dependencies = [
+ "futures-timer",
+ "pin-project",
+]
+
+[[package]]
 name = "fixed-hash"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10409,6 +10419,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4ba7494d250d65dc3439365ac5e8e0fbb9c3992e6e84b7aa01d69e082249b8b8"
 dependencies = [
  "derive-where",
+ "finito",
  "frame-metadata 20.0.0",
  "futures",
  "hex",
@@ -10421,6 +10432,7 @@ dependencies = [
  "subxt-core 0.41.0",
  "subxt-lightclient",
  "thiserror 2.0.12",
+ "tokio",
  "tokio-util",
  "tracing",
  "url",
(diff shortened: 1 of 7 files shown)


kianenigma (approved):
High level lgtm, but I didn't read the whole context. 

I think a lot of these hacks could be due to the current miner's suboptimal design choice. IF I were to write a miner now, I would split it into the following processes, and let them communicate via channels: 

1. `listener`: One that always follows the chain. Either head or finalized.
2. `miner`: One that is always in charge of mining. 

The two can talk to each other. If listeners identifies we are in `Signed`, it will ask `miner` if it is already mining or not. If response is no, it will send it a signal to mine.

PR #1080, opened 2025-06-10
Migrate from RPC calls to ChainHead backend

### LINKED ISSUE #1033: rpc: use `reconnecting-rpc-client` from subxt
This miner currently needs to be restarted if the RPC connection is dropped and
it could be beneficial to automatically reconnect but possible to lose a few blocks.

This chainhead RPC backend will work better for the reconnecting-rpc-client is should used
if this is desired.

The downside is that grafana etc won't detect when the connection is lost without a few tweaks,

Example how to use in subxt can be found here: https://github.com/paritytech/subxt/blob/master/subxt/examples/setup_rpc_chainhead_backend.rs

sigurpol:
This change updates the staking-miner to use the new `ChainHead` backend API instead of direct RPC calls. Key changes:

- Remove RpcClient type and direct RPC method calls
- Use ChainHead backend for block subscriptions and runtime calls
- Update version checking and dry run functionality in the deprecated legacy path
- Get account nonce from storage instead of RPC
- Update error handling for ChainHead responses

Close #1033

### DIFF
--- a/src/commands/legacy/dry_run.rs
+++ b/src/commands/legacy/dry_run.rs
@@ -29,6 +29,7 @@ use crate::{
 };
 use codec::Encode;
 use polkadot_sdk::pallet_election_provider_multi_phase::{MinerConfig, RawSolution};
+use subxt::tx::Payload;
 
 pub async fn dry_run_cmd<T>(client: Client, config: DryRunConfig) -> Result<(), Error>
 where
@@ -90,21 +91,46 @@ where
             account_info.data.frozen,
         );
 
-        let nonce = client
-            .rpc()
-            .system_account_next_index(signer.account_id())
-            .await?;
+        // Get current nonce for the account
+        let account_info = storage
+            .fetch(&runtime::storage().system().account(signer.account_id()))
+            .await?
+            .ok_or(Error::AccountDoesNotExists)?;
+        let nonce = account_info.nonce;
+
         let tx = dynamic::signed_solution(raw_solution)?;
-        let params = ExtrinsicParamsBuilder::new().nonce(nonce).build();
-        let xt = client
+        let params = ExtrinsicParamsBuilder::new().nonce(nonce.into()).build();
+        let _xt = client
             .chain_api()
             .tx()
             .create_signed(&tx, &*signer, params)
             .await?;
-        let dry_run_bytes = client.rpc().dry_run(xt.encoded(), config.at).await?;
-        let dry_run_result = dry_run_bytes.into_dry_run_result()?;
 
-        log::info!(target: LOG_TARGET, "dry-run outcome is {:?}", dry_run_result);
+        let call_data = tx
+            .encode_call_data(&client.chain_api().metadata())
+            .map_err(|e| Error::Other(e.to_string()))?;
+        let target_block = if let Some(at) = config.at {
+            at
+        } else {
+            // Use latest block hash if no specific block is provided
+            client.chain_api().blocks().at_latest().await?.hash()
+        };
+
+        match client
+            .chain_api()
+            .backend()
+            .call("DryRunApi_dry_run_call", Some(&call_data), target_block)
+            .await
+        {
+            Ok(result) => {
+                log::info!(target: LOG_TARGET, "dry-run successful: call returned {} bytes", result.len());
+                log::info!(target: LOG_TARGET, "dry-run outcome: transaction would succeed");
+            }
+            Err(e) => {
+                log::warn!(target: LOG_TARGET, "dry-run failed: {:?}", e);
+                return Err(Error::Other(format!("Dry run failed: {}", e)));
+            }
+        }
     }
 
     Ok(())
--- a/src/dynamic/multi_block.rs
+++ b/src/dynamic/multi_block.rs
@@ -359,10 +359,7 @@ pub(crate) async fn submit<T: MinerConfig + Send + Sync + 'static>(
 
     let mut i = 0;
     let tx_status = loop {
-        let nonce = client
-            .rpc()
-            .system_account_next_index(signer.account_id())
-            .await?;
+        let nonce = get_account_nonce(client.chain_api(), signer.account_id()).await?;
 
         // Register score.
         match submit_inner(
@@ -492,10 +489,7 @@ async fn submit_pages_batch<T: MinerConfig + 'static>(
         });
     }
     let mut txs = FuturesUnordered::new();
-    let mut nonce = client
-        .rpc()
-        .system_account_next_index(signer.account_id())
-        .await?;
+    let mut nonce = get_account_nonce(client.chain_api(), signer.account_id()).await?;
 
     // Collect expected pages before consuming the vector
     let expected_pages: HashSet<u32> = pages_to_submit.iter().map(|(page, _)| *page).collect();
@@ -683,10 +677,7 @@ pub(crate) async fn inner_submit_pages_chunked<T: MinerConfig + 'static>(
 /// Submit a bail transaction to revert incomplete submissions
 pub(crate) async fn bail(client: &Client, signer: &Signer, listen: Listen) -> Result<(), Error> {
     let bail_tx = runtime::tx().multi_block_signed().bail();
-    let nonce = client
-        .rpc()
-        .system_account_next_index(signer.account_id())
-        .await?;
+    let nonce = get_account_nonce(client.chain_api(), signer.account_id()).await?;
     let xt_cfg = ExtrinsicParamsBuilder::default().nonce(nonce).build();
     let xt = client
         .chain_api()
@@ -762,4 +753,18 @@ async fn validate_signed_phase_or_bail(
     }
 
     Ok(true)
+}
+
+async fn get_account_nonce(
+    api: &ChainClient,
+    account_id: &subxt::config::substrate::AccountId32,
+) -> Result<u64, Error> {
+    let account_info = api
+        .storage()
+        .at_latest()
+        .await?
+        .fetch(&runtime::storage().system().account(account_id))
+        .await?
+        .ok_or(Error::AccountDoesNotExists)?;
+    Ok(account_info.nonce.into())
 }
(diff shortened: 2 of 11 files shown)


src/commands/legacy/dry_run.rs
```diff
@@ -90,21 +91,46 @@ where
             account_info.data.frozen,
         );
 
-        let nonce = client
-            .rpc()
-            .system_account_next_index(signer.account_id())
-            .await?;
+        // Get current nonce for the account
```
sigurpol, line 94: Note that we are going to discard the whole legacy path. The code there for dry-run, monitor, info etc but it is not meant to be used in production and will be nuked once Kusama and Polkadot's AHM is completed.

src/dynamic/multi_block.rs
```diff
@@ -482,7 +482,7 @@ async fn submit_pages_batch<T: MinerConfig + 'static>(
 		return Ok(SubmissionResult { failed_pages, submitted_pages: HashSet::new() });
 	}
 	let mut txs = FuturesUnordered::new();
-	let mut nonce = client.rpc().system_account_next_index(signer.account_id()).await?;
+	let mut nonce = get_latest_account_nonce(client.chain_api(), signer.account_id()).await?;
```
jsdw, line 485: This may or may not matter to you here, but IIRC there's a difference here to keep in mind:

- The `system_account_next_index` RPC call gets the next account nonce _taking into account current transactions in the TX pool that haven't made it into blocks yet`
- The `AccountNonceApi_account_nonce` call gets the next account nonce _ignoring any current transactions in the pool_.

One would use the former if they think that other transactions are already in the pool from the same account, but in general since the account should be controlled only by the current "user", the program should always know what the next valid account nonce will be anyway since it's the thing submitting transactions :)

src/dynamic/multi_block.rs
```diff
@@ -734,3 +734,38 @@ async fn validate_signed_phase_or_bail(
...
+
+	// Decode the nonce from the runtime API response
+	let nonce: u32 = nonce_value
+		.to_value()
+		.map_err(|e| Error::Other(format!("Failed to decode nonce value: {}", e)))?
+		.as_u128()
+		.ok_or_else(|| Error::Other("Failed to decode nonce as u128".to_string()))?
+		as u32;
+
+	Ok(nonce as u64)
```
jsdw, line 770: Just wondering why you case to `u32` and then to `u64` :)

src/dynamic/multi_block.rs
```diff
@@ -734,3 +734,38 @@ async fn validate_signed_phase_or_bail(
...
 }
+
+async fn get_latest_account_nonce(
+	api: &ChainClient,
+	account_id: &subxt::config::substrate::AccountId32,
+) -> Result<u64, Error> {
+	let runtime_api_call =
+		subxt::dynamic::runtime_api_call("AccountNonceApi", "account_nonce", vec![
+			Value::from_bytes(account_id),
+		]);
```
jsdw, line 745: You can also use Subxt to give you this more directly and save the code via:

```rust
let nonce = api.tx().account_nonce(account_id).await?
```

sigurpol: Done, ty!

src/dynamic/multi_block.rs
sigurpol: More thoughts on this. When we submit a solution, first we get the nonce via account_nonce() and submit the score. Then we wait for it to be confirmed on chain. Then we try to submit pages in batch, again getting the nonce via account_nonce() now, and for each page we increment manually the nonce itself. And then we verify that each page is successfully submitted.
We are not allowed to submit multiple solutions in parallel.

jsdw: > We might be good with AccountNonceApi_account_nonce but is there an equivalent of system_account_next_index I could eventually use to be more on the safe side?

Some discussion around such a call was here: https://github.com/paritytech/json-rpc-interface-spec/issues/156

TL;DR (or at least my take) is that if you are completely in control of the account being used then incrementing the nonce locally is a robust way to do things, and if transactions using the same account are submitted from different places then they could be connected to different nodes (with different tx pools) anyway, so checking the TX pool won't help.

> Not sure if replacing
> ```
> let params = ExtrinsicParamsBuilder::default().nonce(nonce).build()
> ```
> with
> ```
> let params = ExtrinsicParamsBuilder::default().build();
> ```
> would already delegate nonce mgmt to subxt entirely taking into account pending txs...

Subxt doesn't have any special way to know about the TX pool; it just does the same as `api.tx().account_nonce(account_id).await?` internally to know which nonce to set, if none is given :)

> and for each page we increment manually the nonce itself. And then we verify that each page is successfully submitted. Plus there is the logic to re-tx in case of failure.

If you want to submit multiple transactions at once, then I think manually incrementing the nonce each time is the right approach. The node(s) will then handle the ordering for you (though I am not familiar with the TX pool logic and eg how long it will keep around a TX with a nonce that's in the future for instance etc).

If you want to submit one transaction at a time and wait for the transaction to be finalized before submitting the next (definitely slower but the "safe" default) then you can let Subxt handle the nonce etc and not think about it!

kianenigma: > Then we try to submit pages in batch, again getting the nonce via account_nonce() now, and for each page we increment manually the nonce itself

Reading the flow, I also think that manually incrementing the nonce should be robust here.

src/commands/multi_block/monitor.rs
```diff
@@ -424,6 +426,33 @@ where
...
+	// - Block N: the miner task begins mining a solution, submits the score, waits for it to be
+	//   on-chain, and submits pages, etc. This is a slow operation compared to the speed of block
+	//   production.
+	// - the channel buffer allows one block to be queued while processing another one
+	// - block N+1 gets queued right after N before the miner task starts processing N
+	// - block N+2, N+3, ... , N+M are immediately discarded because the miner task is busy
+	//   (`TrySendError::Full` => backpressure)
+	// - When the miner task completes processing block N, it finally processes block N+1. However,
+	// if several minutes have passed, the RPC node may no longer have the block's state available,
+	// resulting in an "invalid block hash" error. This error is not critical, as block N+1 was
```
jsdw, line 441: Is this something that you ran into when switching to the new chainHead APIs, or just related to staking miner logic? I assume this is because the new RPCs may not keep as many old blocks around?

src/main.rs
```diff
@@ -95,9 +94,21 @@ async fn main() -> Result<(), Error> {
...
-	let runtime_version: RuntimeVersion =
-		client.rpc().state_get_runtime_version(None).await?.into();
-	let chain = opt::Chain::from_str(&runtime_version.spec_name)?;
+
+	// Get full runtime version using ChainHead backend Core_version call
+	let latest_block = client.chain_api().blocks().at_latest().await?;
+	let version_bytes = client
+		.chain_api()
+		.backend()
+		.call("Core_version", None, latest_block.hash())
```
jsdw, line 103: Just as a small nit: I'd generally use the "higher level" APIs over the backend once just to give a more stable interface (though I realise that in this case this is probably simpler, and we don't explicitly discourage using the backend interface directly anywhere..)

src/utils.rs
```diff
@@ -129,14 +129,16 @@ where
...
-			Ok(Some(hash)) => Ok(hash),
-			Ok(None) => Err(Error::Other("Latest block not found".into())),
-			Err(e) => Err(e.into()),
+		Listen::Head => {
+			let block = api.blocks().at_latest().await?;
+			Ok(block.hash())
+		},
+		Listen::Finalized => {
+			let finalized_block_ref = api.backend().latest_finalized_block_ref().await?;
+			Ok(finalized_block_ref.hash())
```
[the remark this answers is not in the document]
sigurpol: ah damn, I assumed that `at_latest()` was returning best blocks.
Now, this is a very important point you have raised.
Currently in the miner:
1. By default we are tracking finalized blocks only 
2. we allow to specify via cli `--listen head` to actually listen to current head

The main question is: do we really need this extra option? Yes, it would be faster but we would increase the risk of the user getting slashed? Is it worth it? It might have been helpful in early testing but now?
This is not something I can decide by myself though. I would be in favor to remove the option `--listen head` and just rely on finalized block. ButI would like to 👂  from @kianenigma as well here.

I would definitely appreciate your help @jsdw if we decide to keep the Listen::Head option around. But let me challenge that, first!

sigurpol:
@jsdw , @kianenigma  now that I've dropped support for best blocks, the PR should be ready for (hopefully 🤞 ) final round of review. Thank you both 🙇

src/main.rs
sigurpol: I've tried 
```
	let version_bytes: Vec<u8> = client
		.chain_api()
		.runtime_api()
		.at_latest()
		.await?
		.call_raw("Core_version", None)
		.await?;

	// Debug: Print the raw bytes to understand the format
	log::debug!(target: LOG_TARGET, "Core_version raw bytes length: {}", version_bytes.len());
	log::debug!(target: LOG_TARGET, "Core_version raw bytes: {:?}", version_bytes);
```
but then 
```
	let runtime_version = if let Ok(version) =
		polkadot_sdk::sp_version::RuntimeVersion::decode(&mut &version_bytes[..])
```
returned an error. 
When I tried to decode the `Core_version` response as a full `RuntimeVersion` structure, it failed because:

1. The raw bytes from `Core_version`: `[115, 116, 97, 107, 105, 110, 103, 45, 97, 115, 121, 110, 99, 45, 112, 97, 114, 97, 99, 104, 97, 105, 110]`
2. These decode to the string: `"staking-async-parachain"`
3. This is just the spec name, not a full `RuntimeVersion` structure

What am I doing wrong?

sigurpol: Done in https://github.com/paritytech/polkadot-staking-miner/pull/1080/commits/983890b726886bd846359a6ea6cb8daca947311e

sigurpol: I am on `0.42.1`. Will keep in mind, ty!

Issue #1077, opened 2025-06-08
Miner is lagging behind the latest blocks in the chain.


sigurpol:
While running the miner for hours / days vs WAH we noticed that the miner starts to process significantly old blocks in comparison with latest block in the chain.
The issue is **NOT** with the submission lock mechanism - that part works correctly. The issue is with the **subscription stream itself**.

What's Actually Happening:

1. **Block Subscription Stream**: The `subscription.next()` call receives blocks sequentially
2. **Slow Processing**: When a block takes minutes to process to mine and submit a solution, the subscription stream **buffers subsequent blocks**
3. **FIFO Queue**: The subscription delivers blocks in FIFO order, so you process **old blocks** even after catching up
4. **No Skip-Ahead**: There's no mechanism to "skip to latest" when the processing is slow

sigurpol:
Closed by #1084

PR #1084, opened 2025-06-11
Split monitor into listener and miner tasks

sigurpol:
## Intro

Implemented a new architecture for the monitor command that splits work between a listener task and miner task communicating via a bounded channel. This provides automatic backpressure when the miner is busy, ensuring fresh blocks are processed without buffering stale work.

The implementation maintains full compatibility with the existing API and configuration while solving the core issue of block subscription buffering that was causing miners to lag behind the chain.

Close
[#1077](https://github.com/paritytech/polkadot-staking-miner/issues/1077)

## The solution

The changes include:

- Split monitor into listener and miner tasks
- Add bounded channel (size=1) for backpressure control
- Implement non-blocking message passing between tasks
- Remove submission locks in favor of single-threaded miner
- Error Handling Strategy:
  - **Listener errors**: All critical - process exits immediately
  - **Miner errors**: Classified as critical or recoverable
    - Critical: Account doesn't exist, invalid metadata, persistent RPC failures, ...
    - Recoverable: Already submitted, wrong phase, temporary mining issues, ...
  - For both miner and listener, **RPC issues** are handled transparently by reconnecting RPC client

### Current design -  Single Loop
| Current Design |
|:---|
| ├── Subscribe to blocks (sequential) |
| ├── For each block → Spawn task |
| ├── Tasks compete for submission lock |
| └── Slow tasks cause subscription buffering |

### New design:

| Listener Task | Miner Task |
|:---|:---|
| ├── Subscribe to blocks | ├── Process one block at a time |
| ├── Fast phase checks | ├── Handle mining operations |
| ├── Check miner availability | ├── Manage submissions |
| ├── Send work to miner | └── Report status back |
| └── Never blocks on slow ops |  |

### DIFF
--- a/src/commands/multi_block/monitor.rs
+++ b/src/commands/multi_block/monitor.rs
@@ -502,11 +502,15 @@ where
         CurrentSubmission::Incomplete(s) => {
             if s.score() == paged_raw_solution.score {
                 // Same score, just submit missing pages
-                let mut missing_pages = Vec::new();
-                for page in s.get_missing_pages() {
-                    let solution = paged_raw_solution.solution_pages[page as usize].clone();
-                    missing_pages.push((page, solution));
-                }
+                let missing_pages: Vec<(u32, T::Solution)> = s
+                    .get_missing_pages()
+                    .map(|page| {
+                        (
+                            page,
+                            paged_raw_solution.solution_pages[page as usize].clone(),
+                        )
+                    })
+                    .collect();
 
                 log::info!(target: LOG_TARGET, "Submitting {} missing pages for existing submission", missing_pages.len());
 


src/commands/multi_block/monitor.rs
```diff
@@ -209,45 +419,38 @@ where
...
+
+    // Update balance
+    let account_info = storage
+        .fetch(&runtime::storage().system().account(signer.account_id()))
+        .await?
+        .ok_or(Error::AccountDoesNotExists)?;
+    prometheus::set_balance(account_info.data.free as f64);
+
+    // Update page length
+    snapshot.set_page_length(n_pages);
```
kianenigma, line 432: I couldn't quite understand why `n_pages` in your `Snapshot` is variable and needs to be set. 

The `<T as MinerConfig>::Pages` that you have is always equal to the onchain page count: https://github.com/krayt78/polkadot-sdk/blob/882ab1a9d89ed83e82a9659bb1c9ddc4847a56b4/substrate/frame/election-provider-multi-block/src/lib.rs#L556

src/commands/multi_block/monitor.rs
```diff
@@ -273,37 +478,41 @@ where
...
             }
-            // Revert previous submission and submit the new one.
+            log::debug!(target: LOG_TARGET, "Reverting previous submission to submit better solution");
             dynamic::bail(&client, &signer, listen).await?;
         }
         CurrentSubmission::Incomplete(s) => {
-            // Submit the missing pages.
             if s.score() == paged_raw_solution.score {
+                // Same score, just submit missing pages
                 let mut missing_pages = Vec::new();
```
kianenigma, line 508:
```suggestion
                let missing_pages = s.get_missing_pages().map(|x| x as usize).map(|p| paged_raw_solution[p].clone()).collect::Vec<_>();
```

just an alternative syntax

src/commands/multi_block/monitor.rs
```diff
@@ -62,16 +122,100 @@ where
...
     let mut subscription = match config.listen {
         Listen::Head => client.chain_api().blocks().subscribe_best().await?,
         Listen::Finalized => client.chain_api().blocks().subscribe_finalized().await?,
     };
-    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel::<Error>();
-    let submit_lock = Arc::new(Mutex::new(()));
-    let snapshot = SharedSnapshot::<T>::new(static_types::Pages::get());
-    let mut pending_tasks: Vec<AbortHandle> = Vec::new();
+
     let mut prev_block_signed_phase = false;
```
kianenigma, line 215: suggestion: if this logic ever becomes too complicated to manage, you can replace your internal phase-management-lite with listening to onchain events, and utilizing `Event::PhaseTransitioned { from, to }` as a canonical way. 

For example: 

```
let events = subxt::events::get_it_somehow_idk();
// each block should have just one phase transition, so we get the last one one defensively. 
let maybe_phase_transition = evetns.filter(|e| e.is_phase_transition()).last();
match maybe_phase_transition { 
    Some(PhaseTranisition { from: Off, to: Snapshot(_) }) => { miner_start_collect_snapshot_pages() },
    Some(PhaseTranisition { from: Snapshot(0), to: Signed }) => { miner_start_minint() },
    Some(PhaseTranisition { from: Signed, to: _ }) => { miner_clear_snapshot() },
    _ => { /* nada */ }
}
```

src/commands/multi_block/monitor.rs
sigurpol: doesn't make sense, since we are just setting the proper value in BlockDetails::new() via `let n_pages = static_types::Pages::get();`. Removed call and mtethod.

src/commands/multi_block/monitor.rs
sigurpol: yeah, I like the functional style more and we also eliminate mutability. Thx, done.

sigurpol:
> LGTM! now that these two are separate tasks with clear roles, we can probably have at least some unit tests for them too?

Definitely something to add, will do in a follow-up PR

PR #1093, opened 2025-06-17
Add min signed phase blocks validation

### LINKED ISSUE #1083: Improve logic to submit score / pages  in signed phase if the miner estimates to have enough  Signed blocks
While testing on WAH, I ended up slashed in the following scenario:

- start running the miner close to the end of Signed phases (e.g. Signed(2))
- miner starts mining, submitting score, verifying the score is on chain, and then submitting pages but it takes time, it's a slow operation in comparison with block production (~6s)
- in the meanwhile we transition to SignedValidation
- at this point the miner within the task associated to processing block in e.g. Signed(32) is not checking the phase while submitting pages
- we might end up in a case where we submitted the score in Signed phase but not all the pages -> the user gets slashed

This can happen during tests and it's a bit more unrealistic for a miner running 24/7. But it's a good check to add in the miner's logic to be on the safer side. 

#1076 has added the logic to verify current (real) phase before submitting score and pages.

This is a good safey net but it's not enough. We need something more robust.

Copy-paste from Element of a thread between @kianenigma  and @sigurpol : 

> @kianenigma :
> If miner is just starting, and sees that we are in Phase::Signed(x), it should only start doing stuff if x > 10 (or any other parameter that we hardcode or make configurable via CLI).
> 10 is basically the minimum duration of the signed phase that we think we need, and if there is less than that left, we don't start as we anticipate we will not finish submitting all pages.
> 
> @sigurpol : 
> agree, the miner also knows how long a Signed phase lasts so the decision will be a function of (how many pages we estimate we need to mine a solution knowing the number of pages a solution is made of,  duration of the signed phase ). But it needs to be slightly smarter than this to handle a restart of the miner (e.g. miner start mining an incomplete solution + crash + restart while still in signed phase -> if we estimate we don't have time to submit the remaining pages, we should bail)
> 
> @kianenigma : 
> I can also think of two options here:
> 
> scenario: Miner starts. It realizes phase is signed. It then checks if we have submitted a solution metadata in the current round.
> 
> Then it has two options:
> 
> 1. smart: see if any page is missing (can easily be checked in the solution metadata pages: Vec<bool>)? if so, re-mine, and submit missing pages
> 2. crude: call bail
> 
> @sigurpol : 
> I was suggesting option 1. , but to be fair, considering it's a rare corner case that should never occur in normal circumstances, option 2 might be sufficient
>

sigurpol:
The changes add validation to ensure there are enough blocks remaining in the signed phase before attempting solution submission. 

This helps prevent incomplete submissions by requiring a minimum number of blocks (default 10) to complete all pages.

The validation checks remaining blocks before starting submission and between page chunks. If there’s insufficient time, it bails any incomplete submissions and skips mining new solutions.

Example:
```bash
./target/release/polkadot-staking-miner --uri ws://127.0.0.1:9946 monitor --seed-or-path //Alice --min-signed-phase-block 5
```

Close #1083.

### DIFF
--- a/src/dynamic/multi_block.rs
+++ b/src/dynamic/multi_block.rs
@@ -570,7 +570,7 @@ async fn submit_pages_batch<T: MinerConfig + 'static>(
 					{
 						let page = solution_stored.2;
 
-						log::info!(
+						log::debug!(
 							target: LOG_TARGET,
 							"Page {page} included in block {:?}",
 							hash
@@ -781,7 +781,7 @@ async fn validate_signed_phase_or_bail(
 					if submitted_pages < n_pages as usize {
 						log::info!(
 							target: LOG_TARGET,
-							"Bailing incomplete submission for round {} ({}/{} pages submitted) due to insufficient time",
+							"Bailing incomplete submission for round {} ({}/{} pages submitted) due to insufficient signed blocks remaining",
 							round,
 							submitted_pages,
 							n_pages
@@ -794,9 +794,9 @@ async fn validate_signed_phase_or_bail(
 
 				Ok(false)
 			} else {
-				log::debug!(
+				log::trace!(
 					target: LOG_TARGET,
-					"Signed phase has {} blocks remaining, sufficient time to continue",
+					"Signed phase has {} blocks remaining, enough blocks to continue",
 					blocks_remaining
 				);
 				Ok(true)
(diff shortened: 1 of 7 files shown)


src/dynamic/multi_block.rs
```diff
@@ -681,6 +743,7 @@ async fn validate_signed_phase_or_bail(
 	signer: &Signer,
 	listen: Listen,
 	round: u32,
+	min_signed_phase_blocks: u32,
```
kianenigma, line 746: I think the function doc is not accurate by now. I would expect it to say: 

> if we are in the signed phase, and we have a partial solution, and less than `min_signed_phase_blocks` are left, we bail.

PR #1227, opened 2026-02-03
Api Functionality for Offline Election Prediction Tool

Ipsa11:
It covers the API functionality for the offline election prediction tool

### DIFF
--- a/src/commands/server.rs
+++ b/src/commands/server.rs
@@ -16,7 +16,7 @@ use crate::{
 	static_types::multi_block::Pages,
 	utils::{encode_account_id, get_block_hash, get_ss58_prefix},
 };
-use http_body_util::{BodyExt, Full};
+use http_body_util::{BodyExt, Full, Limited};
 use hyper::{Method, Request, Response, body::Bytes, header::CONTENT_TYPE, service::service_fn};
 use hyper_util::{
 	rt::{TokioExecutor, TokioIo},
@@ -38,7 +38,8 @@ type Body = Full<Bytes>;
 pub const MAX_BODY_SIZE: u64 = 1024 * 1024;
 pub const MAX_CONCURRENT_CONNECTIONS: usize = 25;
 pub const MAX_CONCURRENT_PREDICTIONS: usize = 1;
-pub const SIMULATE_TIMEOUT: Duration = Duration::from_secs(600);
+pub const SIMULATE_TIMEOUT: Duration = Duration::from_secs(600); // 10 min
+pub const SNAPSHOT_TIMEOUT: Duration = Duration::from_secs(120); // 2 min
 
 /// Abstracts the two data-fetching operations the HTTP server needs.
 pub trait ServerHandler: Send + Sync + Clone + 'static {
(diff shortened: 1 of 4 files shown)


src/commands/server.rs
```diff
@@ -0,0 +1,294 @@
...
+where
+	T: MinerConfig<AccountId = AccountId> + Send + Sync + 'static,
+	T::Solution: Send,
+	T::Pages: Send,
+	T::TargetSnapshotPerBlock: Send,
+	T::VoterSnapshotPerBlock: Send,
+	T::MaxVotesPerVoter: Send,
+{
+	match (req.method(), req.uri().path()) {
+		(&Method::POST, "/simulate") => {
```
sigurpol: do we handle request timeout in case /simulate request hangs (RPC down,s mining stalled etc)? Probably a tokyo timeout is needed here, wdyt?

tests/server.rs
```diff
@@ -0,0 +1,83 @@
+//! Tests for the server command
+
+use clap::Parser;
+use polkadot_staking_miner::commands::types::{
+	ElectionAlgorithm, OverridesConfig, PredictConfig, ServerConfig,
+};
+
+/// Test server command help works
+#[test]
```
sigurpol: I think we are missing the most important tests here. 
No tests actually start the server and make HTTP requests. The server's routing, error responses, and query param handling are untested, we are only testing parsing and serialization

sigurpol (changes requested):
Generally looks good but I think we need some more work before merging.

src/commands/server.rs
```diff
@@ -0,0 +1,294 @@
...
+			let query = req.uri().query().unwrap_or("");
+			let query_block_number =
+				parse_query_param(query, "block").and_then(|s| s.parse::<u32>().ok());
+
+			let body_bytes = match req.collect().await {
```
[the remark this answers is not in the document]
sigurpol: sounds good!

Ipsa11:
Hey @sigurpol,
I’ve updated the entire functionality based on the feedback received. Could you please review the latest changes and let me know if everything works as expected from your side?

src/commands/server.rs
```diff
@@ -0,0 +1,371 @@
...
+
+					json_response(&serde_json::json!({
+						"result": SimulateResult {
+							run_parameters,
+							validators,
+							nominators,
+						}
+					}))
+				},
+				Ok(Err(e)) => {
```
sigurpol: The `/snapshot` endpoint has no timeout. If the RPC hangs or the chain query stalls, this connection sits open indefinitely. Unlike `/simulate` which has `SIMULATE_TIMEOUT`, this path goes straight to `handler.snapshot()` with no guard.

Suggest wrapping it similarly:
```rust
let result = timeout(SNAPSHOT_TIMEOUT, handler.snapshot(query_block_number)).await;
match result {
    Ok(Ok(snapshot_data)) => json_response(&serde_json::json!({ "result": snapshot_data })),
    Ok(Err(e)) => { ... },
    Err(_) => error_response(504, "Gateway Timeout".to_string()),
}
```

src/commands/types.rs
```diff
@@ -117,88 +134,111 @@ pub struct PredictConfig {
 
 	/// Output directory for prediction results
 	#[clap(long, default_value = "results")]
-	pub output_dir: String,
+	#[serde(default)]
```
sigurpol: [nit] `output_dir` is meaningless for API callers and `run_prediction` doesn't use it. Adding `#[serde(skip)]` would keep it CLI-only and prevent confusion in the API contract. (It's harmless as-is since the server path never reads it, but it leaks implementation detail into the API surface.)

sigurpol (changes requested):
Thanks @Ipsa11 — great progress. Most of the original feedback has been addressed :rocket: 

Potential remaining issues:

1. Body size check is ineffective  — `req.collect().await` reads the full body into memory *before* checking size. A multi-GB payload still causes OOM. Use `http_body_util::Limited` to enforce the limit during streaming. 

2. No timeout on `/snapshot`  — `/simulate` has `SIMULATE_TIMEOUT` but `/snapshot` can hang indefinitely if the RPC stalls. Wrap it with a timeout too.

3. `output_dir` in API surface (nit) — Harmless but confusing. `#[serde(skip)]` would clean it up.


[UPDATE] Please also take a look at CI issue here -> https://github.com/paritytech/polkadot-staking-miner/actions/runs/22297072230/job/64499370085?pr=1227

PR #1302, opened 2026-07-29
fix: implement the RestrictOrigins transaction extension

sigurpol:
Paseo Asset Hub carries `pallet-origin-restriction`'s `RestrictOrigins`, whose value is a bool. subxt's `frame-decode` skips extensions with an empty value type and encodes `None` for `Option` ones, but it cannot handle anything different from None or Option like  a bool in this case, so every submission failed at encode time with

```
TransactionExtensions(NotFound("RestrictOrigins"))
```

after mining.

Drop `PolkadotConfig` which was intended for RelayChain only, and instead rely on `SubstrateConfig` in a miner config that adds the `RestrictOrigins` extension, always **enabled**: the `pallet-origin-restriction` rejects a restricted origin when it is disabled, while enabling it only costs pre-dispatch weight for the unrestricted origins the miner signs with.

Chains that do not declare it never ask for it, so Polkadot, Kusama and Westend Asset Hub are currently unaffected.

The code in this PR can be greatly simplified once [subxt #2265](https://github.com/paritytech/subxt/issues/2265) is addressed.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -51,7 +51,7 @@ jobs:
           echo "$HOME/.bun/bin" >> $GITHUB_PATH
 
       - name: Set up Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@v7
         with:
           node-version: "18"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8311,7 +8311,6 @@ dependencies = [
  "prometheus 0.14.0",
  "regex",
  "scale-info",
- "scale-info-legacy",
  "scale-value",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8277,6 +8277,7 @@ dependencies = [
  "prometheus 0.14.0",
  "regex",
  "scale-info",
+ "scale-info-legacy",
  "scale-value",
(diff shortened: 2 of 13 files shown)


sigurpol:
@lexnv @dennisparity  can you please verify that this is a sane way to handle it subxt-wise  before  https://github.com/paritytech/subxt/issues/2265 is fixed (assuming it will be fixed...) ? I am not an expert here so please take a look :bow:

sigurpol:
> @lexnv @dennisparity can you please verify that this is a sane way to handle it subxt-wise before [paritytech/subxt#2265](https://github.com/paritytech/subxt/issues/2265) is fixed (assuming it will be fixed...) ? I am not an expert here so please take a look 🙇

@lexnv  - have you had a chance to look at this PR? It's needed at least as temporary workaround (see subxt 2265) to unlock elections on Paseo AH, thanks!

Ank4n (approved):
Is this compatible with chains that don't have the `RestrictOrigins` extension? Assuming you have verified that - LGTM!

sigurpol:
> Is this compatible with chains that don't have the `RestrictOrigins` extension? Assuming you have verified that - LGTM!

Yes it is compatible with the other chains as well , I have verified against PAH and WAH. Ty

PR #1303, opened 2026-07-29
feat: accept self-stake in predict candidates_include overrides

### LINKED ISSUE #1300: Feature request: Add self-stake to candidates in predict override file
Staking miner has a `predict` command that will predict the election result with or without a snapshot. You can also provide an override file that can contain arbitrary candidates and voters. The format is described [here](https://github.com/paritytech/polkadot-staking-miner#election-overrides-file-format).

The `candidates_include` key is an array of addresses that you want to be considered as candidates and which usually aren't. The tool will consider the bonded amount in the accounts, which usually is zero. However, now that self-stake has a mandatory minimum and can be quite high, in order for `predict` to work properly with `candidates_include` it also needs arbitrary self-stakes. 

So, I'd like to ask that `candidates_include` is changed to accept arrays as values, similar to `voters_include`, where in each array the first argument is the address and the second the self-stake. So, the override format should be like this:
```
{
  "candidates_include": [
    ["15S7YtETM31QxYYqubAwRJKRSM4v4Ua6WGFYnx1VuFBnWqdG", 100000000000000]
  ],
  "candidates_exclude": [],
  "voters_include": [
    ["15S7YtETM31QxYYqubAwRJKRSM4v4Ua6WGFYnx1VuFBnWqdG", 1000000, ["15S7YtETM31QxYYqubAwRJKRSM4v4Ua6WGFYnx1VuFBnWqdG"]]
  ],
  "voters_exclude": []
}
```

sigurpol:
`candidates_include` entries take an optional self-stake as `["addr", stake]`. The target snapshot carries no stake, so a candidate's self-stake only reaches the election through the voter snapshot. Each entry therefore materializes a self-vote, as the chain does for its own validators.

An entry for an account already in the fetched data overrides its self-stake, rewriting only that account's own vote and leaving votes cast into it untouched, which makes "raise validator X's self-stake, keeping its nominators" expressible. `voters_include` is applied afterwards and takes precedence. Bare addresses keep parsing, with no self-stake.

Also cap the voter set at `Pages * VoterSnapshotPerBlock` when overrides exceed it.

Closes #1300.

### DIFF
--- a/src/dynamic/election_data.rs
+++ b/src/dynamic/election_data.rs
@@ -27,7 +27,6 @@ use crate::{
 	dynamic::staking::{fetch_candidates, fetch_voters},
 	error::Error,
 	prelude::{AccountId, AtBlock, LOG_TARGET},
-	static_types::multi_block::VoterSnapshotPerBlock,
 	utils::{encode_account_id, planck_to_token, planck_to_token_u64, read_data_from_json_file},
 };
 
@@ -49,7 +48,7 @@ pub struct PredictionContext<'a> {
 /// Returns a single-page target snapshot and a Vec of voter pages
 pub(crate) fn convert_election_data_to_snapshots<T>(
 	candidates: Vec<ValidatorData>,
-	voters: Vec<NominatorData>,
+	mut voters: Vec<NominatorData>,
 	data_source: ElectionDataSource,
 ) -> Result<(TargetSnapshotPageOf<T>, Vec<VoterSnapshotPageOf<T>>), Error>
 where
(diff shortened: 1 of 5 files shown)


src/dynamic/election_data.rs
```diff
@@ -83,7 +82,34 @@ where
...
-	let per_voter_page = VoterSnapshotPerBlock::get();
+	let per_voter_page = T::VoterSnapshotPerBlock::get();
+
+	// The miner mines at most `T::Pages` pages and bounds the voter pages to that many, so an extra
+	// page is dropped along with every voter in it. Worse, staking data is paginated
+	// strongest-first and then reversed below, which makes the dropped page the strongest one.
+	// Only overrides can exceed the capacity (fetched data is already truncated to it), so let the
+	// weakest voters make room instead, the way the chain's own snapshot keeps the strongest.
+	let capacity = (T::Pages::get() as usize).saturating_mul(per_voter_page as usize);
+	if capacity > 0 && voters.len() > capacity {
```
kianenigma, line 93: as `voters` been already mutated to already include the self-stake by this LoC?

If a validator has some self-stake, that self-stake may or may not make it to the the election voters depending on where it lands on the bags-list, as you said in the PR description. So it would be unrealistic to force-include the self-stake, if that is the case. We should `.sort` all voters including the self-stake, `.take(22500)` again.

Note: With the emphasis of the self-stake, it is reasonable to now change the `get_npos_voters` function in staking to always take the self-stakes first, but since they are all at least 10k, they will almost always be the case.

kianenigma (approved):

src/dynamic/election_data.rs
sigurpol: > as voters been already mutated to already include the self-stake by this LoC?

yes, `apply_overrides` runs before this, so voters already contains the injected self-vote by this line.

> If a validator has some self-stake, that self-stake may or may not make it to the the election voters depending on where it lands on the bags-list, as you said in the PR description. So it would be unrealistic to force-include the self-stake, if that is the case. We should .sort all voters including the self-stake, .take(22500) again.

That's what the trim does: it sorts all voters by stake, self-stake included, and keeps the top capacity so that the injected one competes like any other and is dropped if it doesn't make the cut. 
I have added  a new test covering for this exact case and improved the comment here.

About the note: yes, that would be probably a reasonable SDK change to land.

PR #1325, opened 2026-09-08
workaround(ci): unbreak the nightly netcat install

### LINKED ISSUE #1324: staking-miner integration tests failed against latest polkadot build.
Go to [job logs](https://github.com/paritytech/polkadot-staking-miner/commit/13c00249a156392dc97a7d6a53e268e6eaf6633e/checks/34172769661) to see details about the failure.

sigurpol:
Debian 11 LTS ended `2026-08-31`, `bullseye-security`'s Release file is inow  expired and `apt-get update` fails.

```
Reading package lists...
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired (invalid since 7h 43min 22s). Updates for this repository will not be applied.
```

causing test to fail because of 

```
> /tmp/just-inC80o/setup: line 26: nc: command not found
❌ Timeout waiting for port 9944
```

Until we bump to a new Debian image, skip the `Valid-Until` check in the verify step.

Testing the fix here: https://github.com/paritytech/polkadot-staking-miner/actions/runs/34205052146

```
Preparing to unpack .../netcat-openbsd_1.217-3_amd64.deb ...
Unpacking netcat-openbsd (1.217-3) ...
Setting up netcat-openbsd (1.217-3) ...
update-alternatives: using /bin/nc.openbsd to provide /bin/nc (nc) in auto mode
```

Close https://github.com/paritytech/polkadot-staking-miner/issues/1324.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -48,6 +48,7 @@ jobs:
           # [TEMPORARY WORKAROUND] Debian 11 LTS ended on 2026-08-31, so
           # bullseye-security's Release file is expired and apt fails the whole update
           # over it. Until we bump to a new image, let's temporarily add `Check-Valid-Until=false` flag.
+          # TODO: remove once we have updated to a newer image.
           apt-get -o Acquire::Check-Valid-Until=false update
           apt-get install -y netcat-openbsd
           # Install just
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -41,9 +41,15 @@ jobs:
           path: polkadot-sdk
 
       - name: Install dependencies
+        shell: bash
         run: |
-          # Install netcat for port checking
-          apt-get update && apt-get install -y netcat
+          set -euo pipefail
+          # Install netcat for port checking.
+          # [TEMPORARY WORKAROUND] Debian 11 LTS ended on 2026-08-31, so
+          # bullseye-security's Release file is expired and apt fails the whole update
+          # over it. Until we bump to a new image, let's temporarily add `Check-Valid-Until=false` flag.
+          apt-get -o Acquire::Check-Valid-Until=false update
+          apt-get install -y netcat-openbsd
           # Install just
           curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/local/bin
           # Install bun
@@ -67,6 +73,7 @@ jobs:
           zombienet --help
           just --version
           bun --version
+          command -v nc
 
       - name: Cache Rust dependencies for polkadot-sdk
         uses: swatinem/rust-cache@v2


.github/workflows/nightly.yml
```diff
@@ -41,9 +41,15 @@ jobs:
...
         run: |
-          # Install netcat for port checking
-          apt-get update && apt-get install -y netcat
+          set -euo pipefail
+          # Install netcat for port checking.
+          # [TEMPORARY WORKAROUND] Debian 11 LTS ended on 2026-08-31, so
+          # bullseye-security's Release file is expired and apt fails the whole update
+          # over it. Until we bump to a new image, let's temporarily add `Check-Valid-Until=false` flag.
+          apt-get -o Acquire::Check-Valid-Until=false update
+          apt-get install -y netcat-openbsd
```
sigurpol, line 53: Note: in bullseye:
```
apt-get install -y netcat on bullseye gives you:
The following additional packages will be installed:
  netcat-openbsd
The following NEW packages will be installed:
  netcat netcat-openbsd
```

but starting from newer images (bookworm onward) `netcat` doesn't ship the binary,  whereas netcat-openbsd does.  I would suggest to add `netcat-openbsd` already now so we are ready for when the image is updated.
Alternative: we could get rid of netcat dependency completely in SDK https://github.com/paritytech/polkadot-sdk/blob/fd42d2f58404290bfff77492f4e48c531d209bf4/substrate/frame/staking-async/runtimes/papi-tests/justfile?plain=1#L26
and then we don't require netcat to be installed at all here

cirko33: Is this alternative better? Sometimes less can become more
I agree this solution is temporary, so agreeing on removing netcat.

.github/workflows/nightly.yml
```diff
@@ -41,9 +41,15 @@ jobs:
           path: polkadot-sdk
 
       - name: Install dependencies
+        shell: bash
         run: |
-          # Install netcat for port checking
-          apt-get update && apt-get install -y netcat
+          set -euo pipefail
+          # Install netcat for port checking.
+          # [TEMPORARY WORKAROUND] Debian 11 LTS ended on 2026-08-31, so
```
andreitrand, line 48: Since we're lowering security here, I feel we should be more verbose about restoring it once the Debian update happens. At the very least, I would add a `TODO` label here, so it's easy to find in the future.

andreitrand (approved):

.github/workflows/nightly.yml
sigurpol: the suggested change in upstream SDK's justfile is mostly to replace netcat with some bash magic like ` </dev/tcp/localhost/$port` or similar - so one thing less to install here (the less the better indeed) - can be done, very low prio imho though

.github/workflows/nightly.yml
sigurpol: added a TODO

andreitrand (approved):

